PR #33, opened 2026-03-15
fix: prevent Enter from submitting during IME composition

UnsongK:
Fix IME composition issue with Enter key.
When using CJK input methods(Chinese/Japanese/Korean), pressing Enter to confirm character selection was incorrectly triggering form submission. 
This fix listens for compositionstart/compositionend events and blocks Enter keydown during active IME composition.


lsdefine:
Thanks for the contribution! I've incorporated the IME fix into main with a few changes:
- Compressed the JS to a single line
- Added `if os.name != 'nt':` guard so Windows users are completely unaffected
- Removed the height parameter

Closing this PR. Appreciate the effort! 🙏

PR #77, opened 2026-04-16
Replace deprecated Streamlit HTML injections with st.iframe

wnfxsa1213:
## Summary
- replace deprecated `streamlit.components.v1.html` injections with `st.iframe`
- remove the now-unused `streamlit.components.v1` import from `frontends/stapp2.py`
- use minimal valid iframe dimensions (`height=1`, `width=1` where needed) for hidden script injection

## Testing
- launched `frontends/stapp.py` and `frontends/stapp2.py` under Streamlit 1.56.0
- verified both pages open successfully
- verified the `st.components.v1.html will be removed after 2025-11-01` deprecation warning no longer appears
- verified injected iframe/script content is still present and page loads without new frontend errors

## Notes
- excluded unrelated local `.gitignore` change from this PR

### DIFF
--- a/frontends/stapp.py
+++ b/frontends/stapp.py
@@ -139,7 +139,6 @@ for msg in st.session_state.messages:
 
 # Scroll-height ghost fix: during streaming, expander open/close mid-animation can leave
 # phantom height → scrollbar long but can't scroll to bottom. Periodically detect & reflow.
-import streamlit.components.v1 as components
 _js_scroll_fix = ("!function(){var p=window.parent;if(p.__sfx)return;p.__sfx=1;"
     "var d=p.document;setInterval(function(){"
     "var m=d.querySelector('section.main');if(!m)return;"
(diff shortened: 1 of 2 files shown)


lsdefine:
Thanks for flagging this! You're right that `components.v1.html` is deprecated in Streamlit 1.56+ and `st.iframe` is the official replacement.

I didn't merge this directly because it drops support for Streamlit ≤1.55 (where `st.iframe` doesn't exist). I've gone with a compatibility layer instead:

```python
try:
    from streamlit import iframe as _st_iframe  # 1.56+
    _embed_html = lambda html, **kw: _st_iframe(html, **{k: max(v, 1) if isinstance(v, int) else v for k, v in kw.items()})
except (ImportError, AttributeError):
    from streamlit.components.v1 import html as _embed_html  # ≤1.55
```

Appreciate the contribution 🙏

PR #88, opened 2026-04-17
fix: condense consecutive newlines instead of deleting them in _clean_content

### LINKED ISSUE #87: fix: _clean_content deletes excess newlines instead of condensing them
## Bug in `agent_loop.py` — `_clean_content` function

In `_clean_content`, the following loop is used:

```python
    text = re.sub(p, '\n\n' if '\\\\n' in p else '', text)
```

The condition `'\\\\n' in p` is checking for a **3-character** string (`\\n`: backslash, backslash, n) inside the third pattern `r'(\r?\n){3,}'`, which only contains the **2-character** sequence `\n` (backslash, n).

Because this check is always `False` for the third pattern, 3+ consecutive newlines are **deleted entirely** (replaced with `''`) instead of being **condensed to two newlines** (`'\n\n'`).

This causes all paragraph spacing in LLM responses to disappear in non-verbose mode.

## Fix

Change `'\\\\n'` to `'\\n'` so the condition correctly identifies the newline-collapse pattern:

```python
```

Signed-off-by: Cocoon-Break <54054995+kuishou68@users.noreply.github.com>

kuishou68:
Closes #87

## Problem
In `_clean_content`, the replacement condition uses `'\\\\n' in p` which is a 3-character string (`\\n`: backslash + backslash + n). This never matches the third regex pattern `r'(\r?\n){3,}'` which contains the 2-character sequence `\n`. As a result, groups of 3+ newlines are **deleted** (replaced with `''`) rather than **condensed to two newlines**.

This causes all paragraph spacing to disappear in non-verbose mode LLM output.

## Fix
Changed `'\\\\n' in p` to `'\\n' in p` so the condition correctly matches the newline-condensing pattern.

Signed-off-by: Cocoon-Break <54054995+kuishou68@users.noreply.github.com>


lsdefine (commented):
Hi @kuishou68, thanks for the contribution! I'm GA (GenericAgent) — I reviewed this PR by actually running the code to verify.

### Analysis: The existing code is correct ✅

The key logic (`agent_loop.py` L112-113):
```python
for p in [..., r'(\r?\n){3,}']:
```

The PR description states that `'\\n' in p` is a 3-character string (`\\` + `\\` + `n`) that can never match the 2-character `\n` in the raw string pattern. **This is incorrect.**

In Python source code, `'\\n'` is an escape sequence that produces a **2-character** string: `\` + `n`. The raw string `r'(\r?\n){3,}'` also contains a literal `\` + `n`. So `'\\n' in p` evaluates to **`True`** for the newline pattern, and the replacement is correctly `'\n\n'` (condense to two newlines).

### Experimental verification

```python
>>> p = r'(\r?\n){3,}'
>>> '\\n' in p
True

>>> import re
>>> test = "hello\n\n\n\n\nworld"
>>> re.sub(r'(\r?\n){3,}', '\n\n' if '\\n' in r'(\r?\n){3,}' else '', test)
'hello\n\nworld'   # ✅ Correctly condensed, not deleted
```

### Additionally: this PR has 0 file changes

The PR contains 1 commit but **0 additions, 0 deletions, 0 changed files** — it appears no actual code modification was included.

---

Closing this PR as the existing behavior is already correct. If you believe there's still an issue, feel free to open a new issue with a reproducible test case!

PR #97, opened 2026-04-17
feat(wechat): add wechat_allowed_users gating, mirror QQ adapter pattern

### LINKED ISSUE #95: wechatapp.py has no allowed_users gating, unlike every other adapter
## Observation

Every other chat-app adapter in `frontends/` reads an allow-list from
`mykey.py` and either rejects or logs unknown users:

| Adapter             | Reads allow-list                | Behavior on empty list |
|---------------------|---------------------------------|------------------------|
| `tgapp.py`          | `tg_allowed_users`              | `exit()` (line 122-124) |
| `fsapp.py`          | `fs_allowed_users`              | public if `[*]`         |
| `qqapp.py`          | `qq_allowed_users`              | public if `[*]`         |
| `dingtalkapp.py`    | `dingtalk_allowed_users`        | public if `[*]`         |
| `wecomapp.py`       | `wecom_allowed_users`           | public if `[*]`         |
| **`wechatapp.py`**  | **(none — keyword grep returns 0 matches)** | n/a — no gate at all |

`README.md` (Chinese half, line 297) tells the user the wechat adapter
needs no extra config: \"扫码登录即可\". Today that means after the QR
login, any contact who messages the bound WeChat account drives the
agent — `code_run`, `file_read`, `file_write`, etc. all reachable.

This looks like a documentation/implementation gap rather than a design
choice, since the same author wrote the Telegram adapter that hard-exits
on an empty allow-list.

## Proposal (would PR after maintainer alignment)

1. Add `wechat_allowed_users = []` example to `mykey_template.py` with
   a short comment matching the QQ/DingTalk pattern.
2. In `wechatapp.py` message dispatch, drop messages whose sender openid
   isn't in the allow-list. Mirror the Telegram pattern (`exit()` on
   empty list) since this adapter has no \"public webhook\" use case
   — every user is talking to the maintainer's personal WeChat.

Two questions before I send a PR:

- Do you want the same `[\"*\"]` opt-in escape hatch the other adapters
  have, or follow Telegram's stricter \"non-empty required\" rule?
- WeChat senders are identified by `openid` (or the wxid surface that
  the iLink bot exposes) — should the key match the rest of the
  ecosystem (`wechat_allowed_users = [\"<openid>\", ...]`)?

Happy to align before opening a PR so the change matches your existing
adapter convention.

shaun0927:
Closes #95.

## Problem

`frontends/wechatapp.py` is the only chat adapter that does not look at
an allow-list. Every other adapter under `frontends/` reads
`<platform>_allowed_users` from `mykey.py` and rejects unknown senders:

| Adapter             | Allow-list key                       |
|---------------------|--------------------------------------|
| `tgapp.py`          | `tg_allowed_users` (non-empty required) |
| `fsapp.py`          | `fs_allowed_users`                   |
| `qqapp.py`          | `qq_allowed_users`                   |
| `dingtalkapp.py`    | `dingtalk_allowed_users`             |
| `wecomapp.py`       | `wecom_allowed_users`                |
| **`wechatapp.py`**  | **(none)**                           |

After `python frontends/wechatapp.py` scans the QR code, anyone who
messages the bound personal WeChat account drives the agent — every
tool the agent has (`code_run`, `file_read`, `file_write`, web control)
is reachable to any contact, with no logging or rejection path.

## Fix

Mirror the QQ adapter pattern that PR #25 established:

```python
ALLOWED = {str(x).strip() for x in mykeys.get('wechat_allowed_users', []) if str(x).strip()}
...
if not public_access(ALLOWED) and uid not in ALLOWED:
    print(f'[WX] unauthorized user: {uid}', file=sys.__stdout__)
    return
```

Three deliberate decisions:

1. **Reuse `chatapp_common.public_access()`** rather than inlining
   `not allowed or '*' in allowed`, so the semantics stay locked to
   the same definition the other adapters use. If `public_access`
   ever changes, wechat picks it up for free.
2. **Default behavior is unchanged** — existing setups that don't set
   `wechat_allowed_users` get `ALLOWED == set()`, `public_access()`
   returns True, and every message still reaches the agent. This
   matches `qq`/`dingtalk`/`wecom`/`fs` exactly. (Telegram is the
   stricter exception that hard-fails on empty; not adopted here to
   avoid breaking existing wechat installs without warning.)
3. **Log line format** matches `qq`/`dingtalk`: `[WX] unauthorized
   user: <uid>`, so a user grepping `temp/wechatapp.log` sees the
   same shape as the other adapters' logs.

`mykey_template.py` gets the matching `# wechat_allowed_users = ['from_user_id']`
comment line in the existing chat-platform block so the option is
discoverable.

## Verification

- `python -m py_compile frontends/wechatapp.py mykey_template.py` — passes.
- Existing wechat installs (no `wechat_allowed_users` in mykey) keep working
  unchanged because `ALLOWED` is empty → `public_access(ALLOWED)` is True
  → the new check is skipped.
- New installs that opt in via `wechat_allowed_users = ['<from_user_id>']`
  get the same allow-list semantics as `qqapp.py` (uid must be in the set).

Diff size: +8 / -0 across 2 files.

### DIFF
--- a/frontends/wechatapp.py
+++ b/frontends/wechatapp.py
@@ -6,6 +6,10 @@ from Crypto.Cipher import AES
 sys.path.insert(0, os.path.dirname(os.path.dirname(os.path.abspath(__file__))))
 _TEMP_DIR = os.path.join(os.path.dirname(os.path.dirname(os.path.abspath(__file__))), 'temp')
 from agentmain import GeneraticAgent
+from llmcore import mykeys
+from chatapp_common import public_access
+
+ALLOWED = {str(x).strip() for x in mykeys.get('wechat_allowed_users', []) if str(x).strip()}
 
 # ── WxBotClient (inline from wx_bot_client.py) ──
 API = 'https://ilinkai.weixin.qq.com'
@@ -233,6 +237,9 @@ def on_message(bot, msg):
     text = bot.extract_text(msg).strip()
     uid = msg.get('from_user_id', '')
     ctx = msg.get('context_token', '')
+    if not public_access(ALLOWED) and uid not in ALLOWED:
+        print(f'[WX] unauthorized user: {uid}', file=sys.__stdout__)
+        return
     media_paths = _dl_media(msg.get('item_list', []))
     if not text and not media_paths: return
     if media_paths:
--- a/mykey_template.py
+++ b/mykey_template.py
@@ -392,6 +392,7 @@ native_oai_config = {
 # wecom_secret = 'your_bot_secret'
 # wecom_allowed_users = ['your_user_id']            # 留空或 ['*'] 表示允许所有企业微信用户
 # wecom_welcome_message = '你好，我在线上。'
+# wechat_allowed_users = ['from_user_id']           # 留空或 ['*'] 表示允许所有微信用户
 # dingtalk_client_id = 'your_app_key'
 # dingtalk_client_secret = 'your_app_secret'
 # dingtalk_allowed_users = ['your_staff_id']        # 留空或 ['*'] 表示允许所有钉钉用户


lsdefine:
Closing. The wechat frontend is a local hook (not a public-facing bot like tg/qq), so the threat model is different — incoming messages are already gated by the user's own WeChat contacts. If access control is needed it'll be handled differently from the tg/qq pattern.

PR #98, opened 2026-04-17
fix: close file handles via context manager in remaining call sites

### LINKED ISSUE #96: Several open() calls without context manager — possible FD leak in long-running bot daemons
## Background

PR #76 fixed one `open(...).read()` instance in `reflect/scheduler.py` by
wrapping it in `with`. Quick grep over the rest of the codebase finds
several similar patterns that rely on garbage collection to close the
handle:

```
ga.py:23     if os.path.exists(cr_header): tmp_file.write(open(cr_header, encoding='utf-8').read())
ga.py:387    old = open(path, 'r', encoding=\"utf-8\").read() if os.path.exists(path) else \"\"
ga.py:388    open(path, 'w', encoding=\"utf-8\").write(new + old)
agentmain.py:16   open(...).read()      # tools_schema load
agentmain.py:24   open(mem_txt, 'w').write(...)
agentmain.py:28   open(...).write(open(...).read())   # nested, two handles
agentmain.py:236  open(...).<append>     # log file
frontends/wechatapp.py:183  open(p, 'wb').write(pt)
```

For the launchable Streamlit/CLI flow these eventually get GC'd and the
process exits anyway, so the existing behavior is mostly fine. But
`tgapp.py`, `wechatapp.py`, `fsapp.py` etc. run as long-lived daemons,
and several of these patterns are inside per-message handlers. On
default Linux ulimit (1024 FDs) a busy bot will eventually trip
`OSError: [Errno 24] Too many open files`.

This may also be the root cause of #48 (\"multiple calls then model
unresponsive, needs restart\") — the symptom there matches FD
exhaustion: the agent stops responding mid-conversation and a restart
clears it.

## Proposal

Single janitor-style PR that converts all of the above to
`with open(...) as f:`. Net line count is +0 (extra `with` line is
balanced by removing the chained call). Same byte intent, no behavior
change for the happy path.

Two questions before I send the PR:

1. Do you prefer a single janitor PR touching ~5 files, or split per
   file? CONTRIBUTING.md asks for \"small change radius\" — I read this
   as \"don't ripple A → B → C\", which the FD-hygiene change does not
   do (each file is independent). But happy to split if you prefer.

2. There are also a few `open(...).read()` calls inside helper closures
   that get called once per process startup (e.g. `_load_mykeys` in
   `llmcore.py`). Should those stay as-is to keep the boot path
   compact, or get the same `with` treatment for consistency?

Reference: PR #76 already established the `with`-context pattern as
acceptable in this codebase, so this is a follow-up applying the same
pattern uniformly.

shaun0927:
Closes #96.

## Problem

PR #76 wrapped the scheduler's `open(...).read()` in a context manager.
Several similar patterns elsewhere still rely on garbage collection to
release the underlying file descriptor. For long-running daemons
(`tgapp.py`, `wechatapp.py`, `fsapp.py`, etc.) the FD eventually leaks
to the per-process limit and the symptom shown in #48 (\"multiple
calls then model unresponsive, needs restart\") becomes hard to
diagnose because the actual `EMFILE` error surfaces in some unrelated
later open.

The recent refactor 47f106c also added a new unmanaged read at
`agentmain.py:105` (`/session.<key> = <file>`), which would have grown
the leak surface.

## Fix

Convert the remaining unmanaged `open(...).<read|write>(...)` patterns
to `with open(...) as f:` blocks. Same modes, encodings, bytes; only the
release timing changes.

| File                       | Lines (pre-patch) | What                          |
|----------------------------|-------------------|-------------------------------|
| `ga.py`                    | 23                | code_run header copy          |
| `ga.py`                    | 387, 388          | prepend mode in `do_file_write` |
| `ga.py`                    | 424               | `_check_plan_completion` read |
| `agentmain.py`             | 16                | tools_schema load             |
| `agentmain.py`             | 24                | global_mem.txt seed           |
| `agentmain.py`             | 27-28             | global_mem_insight.txt seed (nested open().write(open().read()) — 2 leaks) |
| `agentmain.py`             | 33                | tmwebdriver cdp_bridge config seed |
| `agentmain.py`             | 105               | `/session.<key>=<file>` (added by 47f106c) |
| `agentmain.py`             | 242               | reflect log append            |
| `frontends/wechatapp.py`   | 183               | decrypted media write to temp |

## Verification

```
$ python -m py_compile ga.py agentmain.py frontends/wechatapp.py
$ grep -nE 'open\\([^)]*\\)\\s*\\.(read|write)\\(' ga.py agentmain.py frontends/wechatapp.py | grep -v 'with '
(empty — all sites now use a context manager)
```

## Scope choices

- **Single janitor PR** rather than splitting per file. The change is
  the same idiomatic conversion at every site, and reviewing them in
  one place makes it easy to confirm no behavior change.
- **subprocess.Popen(stdout=open(...), stderr=open(...))** at
  `agentmain.py:187-188` is intentionally **not** changed — those file
  objects are owned by the subprocess for its lifetime and closing them
  in the parent breaks the redirection.
- **`if sys.stdout is None: sys.stdout = open(os.devnull, ...)`**
  at lines 3 and 5 is also unchanged because `sys.stdout` keeps the
  reference alive for the entire process.
- **`open(_logf, 'a', ...)` at the wechatapp `__main__` block** is
  unchanged for the same reason — the file is bound to `sys.stdout`
  for the lifetime of the daemon.

Net diff: +21 / -11 lines, no functional change.

### DIFF
--- a/agentmain.py
+++ b/agentmain.py
@@ -13,7 +13,7 @@ from ga import GenericAgentHandler, smart_format, get_global_memory, format_erro
 script_dir = os.path.dirname(os.path.abspath(__file__))
 def load_tool_schema(suffix=''):
     global TOOLS_SCHEMA
-    TS = open(os.path.join(script_dir, f'assets/tools_schema{suffix}.json'), 'r', encoding='utf-8').read()
+    with open(os.path.join(script_dir, f'assets/tools_schema{suffix}.json'), 'r', encoding='utf-8') as f: TS = f.read()
     TOOLS_SCHEMA = json.loads(TS if os.name == 'nt' else TS.replace('powershell', 'bash'))
 load_tool_schema()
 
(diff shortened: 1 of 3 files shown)


lsdefine:
Closing. CPython's reference counting releases these file handles immediately, so this isn't a real resource leak in practice. Prefer to avoid style-only diffs that touch multiple files without fixing an actual bug. Thanks for the thought though.

PR #127, opened 2026-04-21
feat: add auto-follow supervisor with auto-stop

ChengTu-Lazy:
## Summary
- add UI-level auto-follow that can automatically reply to ask_user prompts and continue after a turn completes
- add "auto-stop on done" so the supervisor can disable auto-follow when the task is complete
- preserve structured ask_user payloads and use assistant revision tracking to avoid repeated-text dedupe bugs
## Validation
- python -m py_compile agentmain.py frontends/stapp.py launch.pyw self_supervisor.py
- verified ask_user extraction with both structured loop_result and rendered tool-log samples

### DIFF
--- a/agentmain.py
+++ b/agentmain.py
@@ -39,29 +39,32 @@ def get_system_prompt():
     prompt += get_global_memory()
     return prompt
 
+def build_llm_clients():
+    from llmcore import mykeys
+    llm_sessions = []
+    for k, cfg in mykeys.items():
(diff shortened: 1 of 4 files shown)


lsdefine:
Closing — +454 lines adding a new supervisor module. Per CONTRIBUTING.md skill contribution guidelines: domain-specific workflow modules are marketplace candidates, not core PRs. Also violates "Small change radius" principle.

PR #131, opened 2026-04-22
feat(fsapp): render agent run as one continuously-updating Feishu card

wjl2023:
## Summary

Rework the Feishu (fsapp) frontend so that each run renders as **one continuously-updating task card**, instead of a stream of fragmented messages.

## Motivation

The previous `dq`-based streaming path produced many small message bubbles per turn (one per token batch), which was noisy in group chats and hard to scan after the fact. A single collapsible card gives users a stable anchor: header shows live status, each step is a foldable panel with summary visible and details on demand.

## What changed

All changes are in `frontends/fsapp.py`.

- **Replace `dq` streaming with a pet-style hook** — one `_TaskCard` per user turn; the hook reacts to `summary` / `exit_reason` events from the agent loop and `patch`es the same card.
- **Per-step collapsible panels** — header shows `Turn N · <summary>`; expanding reveals:
  - `💭 Thinking` — model's reasoning (from `response.thinking`, a separate field on `MockResponse`, not inside `content`)
  - `🛠 Tool Calls` — tool name + truncated JSON args
  - `📝 Output` — `response.content` after stripping protocol tags (`<summary>`, `<thinking>`, `<tool_use>`, `<file_content>`) so the header summary isn't duplicated inside the panel
- **Final reply rendered as schema 2.0 markdown card** — consistent visual language with intermediate steps.
- **Code-review pass** (per repo's `code_review_principles.md`):
  - `_TaskCard` owns only stateful card lifecycle (start / step / done / fail).
  - Pure formatting extracted to module-level functions `_build_step_detail(resp, tool_calls)` and `_fmt_tool_call(tc)` — no longer hidden as a staticmethod the hook had to reach into via `card._private`.
  - Hook is now a minimal dispatcher (~10 lines): `exit_reason → card.done`, `summary → card.step`.
  - Flattened a 4-level-nested lambda into a readable named function.

## Testing

Manually tested end-to-end in Feishu across multiple turns:
- Header status updates live (`🤔 思考中 → ✅ 完成`).
- Each step panel expands/collapses independently; all three sections render only when non-empty.
- Final reply renders correctly with markdown.
- Long details are truncated at 8000 chars to stay within card content limits.

### DIFF
--- a/frontends/fsapp.py
+++ b/frontends/fsapp.py
@@ -8,6 +8,7 @@ from frontends.chatapp_common import format_restore
 from frontends.continue_cmd import handle_frontend_command as handle_continue_frontend, reset_conversation
 from llmcore import mykeys
 
+import traceback
 import lark_oapi as lark
 from lark_oapi.api.im.v1 import *
 
@@ -225,6 +226,7 @@ APP_ID = str(mykeys.get("fs_app_id", "") or "").strip()
 APP_SECRET = str(mykeys.get("fs_app_secret", "") or "").strip()
 ALLOWED_USERS = _to_allowed_set(mykeys.get("fs_allowed_users", []))
 PUBLIC_ACCESS = not ALLOWED_USERS or "*" in ALLOWED_USERS
+AGENT_TIMEOUT_SEC = 900
 
 agent = GeneraticAgent()
 threading.Thread(target=agent.run, daemon=True).start()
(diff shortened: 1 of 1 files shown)


lsdefine (approved):
LGTM — 只改了 fsapp.py，不影响其他前端，通过。

PR #135, opened 2026-04-22
feat(tgapp): stream MarkdownV2 replies with draft fallback and file segments

YooooEX:
## Context

  Telegram currently streams replies through a polling + HTML path, which makes draft updates, MarkdownV2 rendering, long-message splitting, and attachment delivery inconsistent with the newer chat frontend behavior.
  This PR keeps the change radius inside `frontends/tgapp.py` plus focused regression tests. No new dependencies are added.

  ## Summary

  Upgrade Telegram reply streaming to use MarkdownV2-first draft/live updates, segmented final output, and attachment-aware delivery.

  ## Changes

  ### `frontends/tgapp.py`
  - switch Telegram streaming from fixed-interval polling edits to event-driven queue consumption
  - prefer `reply_text_draft(...)` in private chats and fall back to normal reply/edit flows when draft updates are unavailable
  - render replies with MarkdownV2 escaping instead of the previous HTML conversion path
  - split long final replies into multiple visible segments instead of truncating or restarting from the tail
  - strip `[FILE:...]` hints from visible text and send extracted files as Telegram attachments after finalize
  - keep stop/error flows explicit by finalizing the current stream session with a notice

### DIFF
--- a/frontends/tgapp.py
+++ b/frontends/tgapp.py
@@ -1,91 +1,275 @@
-import os, sys, re, threading, asyncio, queue as Q, socket, time
+import os, sys, re, threading, asyncio, queue as Q, socket, time, random
 sys.path.insert(0, os.path.dirname(os.path.dirname(os.path.abspath(__file__))))
 _TEMP_DIR = os.path.join(os.path.dirname(os.path.dirname(os.path.abspath(__file__))), 'temp')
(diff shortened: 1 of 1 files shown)


lsdefine (approved):
LGTM

PR #143, opened 2026-04-23
feat: add uv project configuration files

machenme:
Add .python-version, pyproject.toml, and uv.lock to support dependency management via uv (https://docs.astral.sh/uv/).

- .python-version: pins Python 3.12
- pyproject.toml: declares project metadata and dependencies
- uv.lock: lockfile for reproducible installs


lsdefine:
Thanks for the contribution!

We have already handled uv / pyproject support in a lighter way. I don’t think we want to commit a full `uv.lock` or pin the project to Python 3.12 at this stage.

Closing this PR as superseded.

PR #158, opened 2026-04-24
feat: add standalone agent health check script

huangrichao2020:
We received your feedback on the previous PR, and it was reasonable. I reworked the change accordingly and kept this follow-up much smaller and more focused.

This PR only includes a standalone audit helper, without modifying core runtime behavior.

Scope in this PR:
- add a standalone `scripts/agent_health_check.py`
- add tests for the standalone script
- add minimal README usage notes

Explicitly not included:
- no monkey-patching in `ga.py`
- no tool registration in `assets/tools_schema.json`
- no new or modified runtime memory files under `memory/`
- no unrelated macOS installer changes

## Why a standalone third-party style matters

My intention here is to keep `agchk` usable as an external package or helper, rather than making it part of GenericAgent's core runtime.

That means users can adopt it in low-risk ways:
- install `agchk` separately and run it from the terminal
- call the standalone script manually when they want an audit
- ask the agent to run the script or shell command on demand

Instead of binding it into the agent loop, tools schema, or memory files, the idea is to keep it as an optional auditor.

## How users would use it

Direct usage from the terminal:

```bash
python scripts/agent_health_check.py --target-dir /path/to/GenericAgent
python scripts/agent_health_check.py --mode tools --json
```

Typical day-to-day requests a user might give in chat:
- "Run a health check on the current GenericAgent checkout."
- "Audit only the tools layer and show me the highest-severity findings."
- "Check whether memory or prompt layers look bloated or duplicated."
- "Run the standalone audit script and summarize only the actionable issues."

So the intended pattern is:
- user asks for an audit in normal conversation
- the agent runs the standalone script or terminal command
- the audit stays external to core runtime behavior

## Validation

Validation run:
- `python3 -m py_compile scripts/agent_health_check.py`
- `uv run --with pytest python -m pytest tests/test_agent_health_check.py -q`
- `python3 scripts/agent_health_check.py --target-dir /Users/tingchim2pro/Desktop/GenericAgent --mode tools --json`

Note: I also checked a broader local test run and hit pre-existing MiniMax test failures unrelated to this script, so I kept the verification focused on the new standalone helper.

### DIFF
--- a/README.md
+++ b/README.md
@@ -138,6 +138,22 @@ The default Streamlit desktop UI started by `python launch.pyw`, plus the QQ / T
 - `/continue` - list recoverable conversation snapshots
 - `/continue N` - restore the `N`th recoverable conversation
 
+### Standalone Health Check Script
+
+If you want to audit a GenericAgent checkout without modifying runtime behavior, use the standalone helper:
+
+```bash
+python scripts/agent_health_check.py --target-dir /path/to/GenericAgent
+python scripts/agent_health_check.py --mode tools --json
+```
+
+This script is intentionally standalone:
+
+- no monkey-patching
+- no tool registration
+- no direct memory file writes
+- read-only checks against prompts, config, memory layout, tools, loop guards, and hidden layers
+
 
 ## 📊 Comparison with Similar Tools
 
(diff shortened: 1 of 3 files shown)


huangrichao2020:
We received your feedback on the previous PR and reworked this follow-up accordingly. The scope is now intentionally limited to a standalone audit helper plus tests/docs, without touching core runtime integration, runtime memory files, or unrelated installer changes.\n\nThe intended usage is also external-first: users can run the script manually, or in day-to-day conversation ask the agent to run the standalone audit command and summarize the result. In other words, the audit stays outside GenericAgent's core runtime path rather than becoming a built-in tool.

lsdefine:
Closing — +603 lines for a new health-check module violates the small change radius and net-line-count guidelines. See CONTRIBUTING.md.

PR #160, opened 2026-04-24
fix(tgapp): polish Telegram MarkdownV2 rendering and startup helpers

YooooEX:
## Context

Telegram replies already use a MarkdownV2 rendering path, but a few edge cases still made the output less
polished:

- `[FILE:...]` markers were removed from the message body after attachment extraction, so users could receive
attachments without an inline filename reference.
- Telegram startup still duplicated helper logic that is already shared by other chat frontends.
- Common Markdown styles like `__bold__` and `~~strikethrough~~` were not converted into Telegram MarkdownV2
entities.

This PR keeps the change radius inside `frontends/tgapp.py`. No new dependencies are added.

## Summary

Polish Telegram message rendering and startup wiring by reusing shared frontend helpers, keeping file marker
basenames visible in replies, and expanding Markdown-to-MarkdownV2 conversion for common bold and strikethrough
styles.

### `frontends/tgapp.py`

- Reuse shared chat frontend helpers for single-instance locking, runtime checks, and log redirection.
- Keep `[FILE:...]` markers visible in the reply body as basenames while still resolving and sending matching
files as Telegram attachments.
- Convert common Markdown `__bold__` input to Telegram MarkdownV2 bold output.
- Convert GFM-style `~~strikethrough~~` input to Telegram MarkdownV2 strikethrough output.

### DIFF
--- a/frontends/tgapp.py
+++ b/frontends/tgapp.py
@@ -1,4 +1,4 @@
-import os, sys, re, threading, asyncio, queue as Q, socket, time, random
+import os, sys, re, threading, asyncio, queue as Q, time, random
 sys.path.insert(0, os.path.dirname(os.path.dirname(os.path.abspath(__file__))))
 _TEMP_DIR = os.path.join(os.path.dirname(os.path.dirname(os.path.abspath(__file__))), 'temp')
 from agentmain import GeneraticAgent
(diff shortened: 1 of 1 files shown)


lsdefine (approved):
LGTM. Confirmed it applies cleanly and the Telegram polish changes are scoped.

PR #170, opened 2026-04-25
Fix OpenAI max token parameter selection

auston314:
I got some error when I was using gpt-5.4 model.


lsdefine:
Thanks for the PR! I implemented a smaller equivalent fix directly on main in 08181be, covering Responses API max_output_tokens and Chat Completions max_completion_tokens for GPT-5/o-series while keeping max_tokens for other compatible providers. Closing this PR to avoid the larger helper change, but really appreciate the catch and contribution!

PR #177, opened 2026-04-25
Fix Telegram startup without a configured proxy

### LINKED ISSUE #175: Proxy issue for Telegram connection
I was trying to setup connection between my Telegram with GA from my iphone, but I got network issue, and not able to talk.

auston314:
Only use `mykeys['proxy']` when it is explicitly set. This avoids forcing the dead default proxy `http://127.0.0.1:2082` and restores Telegram startup.

Closes #175


lsdefine (approved):
LGTM. The issue is that `proxy=None` still gets passed to `HTTPXRequest`, which can cause connection hangs. Conditionally omitting the `proxy` kwarg is the correct fix.

auston314:
Great. Thanks for your feedback!

PR #187, opened 2026-04-26
Update English key template examples

auston314:
Update the English `mykey.py` template with clearer configuration examples:

- Add tested OpenRouter configuration sample.
- Clarify Telegram bot configuration placeholders.
- Add a note that WeChat bot integration is not supported for North America users, recommending Telegram or another supported channel instead.
- Keep the template valid Python (`py_compile` passed).


lsdefine:
Thanks for the contribution. These files change frequently and small patches tend to conflict with ongoing work. Holding off for now.

PR #256, opened 2026-05-02
fix: mutable default list in code_run stop_signal parameter

Kailigithub:
## Summary

Fix a common Python pitfall where `stop_signal=[]` is used as a mutable default argument in the `code_run()` function signature.

## Problem

In `ga.py:12`, `code_run()` has `stop_signal=[]` as a default parameter. In Python, mutable default arguments are shared across all calls, which means if one call mutates the list, subsequent calls will see the modified default. This is a well-known Python anti-pattern.

## Fix

- Changed default from `stop_signal=[]` to `stop_signal=None`
- Added `stop_signal = stop_signal or []` at the start of the function body
- Behavior is identical for callers (still accepts a list), but the default is now safely immutable

## Testing

- `python3 -m py_compile ga.py` passes
- No functional change — callers either pass an explicit list or get a fresh empty list

### DIFF
--- a/ga.py
+++ b/ga.py
@@ -9,14 +9,14 @@ sys.path.append(os.path.abspath(os.path.join(os.path.dirname(__file__), '..')))
 from agent_loop import BaseHandler, StepOutcome, json_default
 script_dir = os.path.dirname(os.path.abspath(__file__))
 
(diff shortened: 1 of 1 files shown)


lsdefine:

PR #264, opened 2026-05-03
fix: Ubuntu/Linux pip config, requirements.txt, and cross-platform launcher

simer11-jing:
## Summary

- Add `#!/usr/bin/env python3` shebangs to all `.pyw` files for direct execution on Linux
- Fix `launch.pyw` `get_screen_width()` to check `os.name` before accessing `ctypes.windll`
- Create `requirements.txt` with complete categorized dependencies (core, ui, bots, ml, misc)
- Expand `pyproject.toml` optional-dependencies: rename `all-frontends` → `bots`, add `ml` group, add `all` group, add missing `Pillow`/`PySide6` to `ui`
- Add `setup_ubuntu.sh` to install Ubuntu system-level dependencies (GTK, WebKit, python3-gi, etc.)

## Test plan

- [x] Shebangs allow direct execution on Ubuntu: `./launch.pyw`, `./hub.pyw`
- [x] `get_screen_width()` no longer relies on try/except for `ctypes.windll` on Linux
- [x] `requirements.txt` includes all packages referenced by project imports
- [x] `setup_ubuntu.sh` installs all system deps needed by pywebview/PySide6 on Ubuntu 24.04

### DIFF
--- a/frontends/desktop_pet.pyw
+++ b/frontends/desktop_pet.pyw
@@ -1,3 +1,4 @@
+#!/usr/bin/env python3
 """Desktop Pet with HTTP Toast — ~90 lines"""
 import tkinter as tk, threading, random, os, sys
 from http.server import HTTPServer, BaseHTTPRequestHandler
--- a/frontends/desktop_pet_v2.pyw
+++ b/frontends/desktop_pet_v2.pyw
@@ -1,3 +1,4 @@
+#!/usr/bin/env python3
 """Desktop Pet with Skin System — Cross-platform with True Transparency"""
 import os, re, sys, json, threading, io
 from http.server import HTTPServer, BaseHTTPRequestHandler
--- a/hub.pyw
+++ b/hub.pyw
@@ -1,3 +1,4 @@
+#!/usr/bin/env python3
 # launcher.pyw - GenericAgent 服务启动器
 # 纯 tkinter + 标准库，零第三方依赖，跨平台
 import os, sys, socket, subprocess, threading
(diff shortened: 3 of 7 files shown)


lsdefine:
Thanks for the effort on cross-platform support!

Unfortunately I can't merge this PR for the following reasons:

1. **Shebang lines on `.pyw` files**: These files are run via `python xx.pyw`, not executed directly. Shebangs are unnecessary here.

2. **`get_screen_width()` fix**: The call site already has `if os.name == 'nt'` branching (line 135), so the Linux guard inside the function is redundant.

3. **`pyproject.toml` changes** (main blocker):
   - Renaming `all-frontends` → `bots` is a **breaking change** for existing users.
   - Adding `PySide6` (~200MB) and `Pillow` to `[ui]` — the project doesn't use PySide6 at all, and Pillow is only used in one bot frontend, not the UI layer.
   - New `ml`/`all` groups add packages (`ultralytics`, `yara-python`, `langfuse`, `GitPython`, `APScheduler`, etc.) that the project never imports.
   - The extras should reflect **actual project imports**, not speculative dependencies.

4. **`requirements.txt`**: With `pyproject.toml` already defining deps, a separate requirements.txt creates a duplicate maintenance burden.

5. **`setup_ubuntu.sh`**: Most of the system packages listed are only needed for PySide6, which shouldn't be a dependency in the first place.

Appreciate the intent — if you'd like to contribute, smaller focused PRs (e.g. fixing a specific Linux runtime bug with a reproducer) would be much easier to review and merge. Thanks!

PR #267, opened 2026-05-04
feat: file_write auto-resume on truncation

tianxin-w:
## Problem

When `file_write` is used on large files, the LLM's response can hit `max_tokens` 
and get truncated mid-`<file_content>`. Currently this either:
- writes a corrupted/truncated file silently, or
- fails the tool call, requiring manual splitting (the tool desc even warns 
  "HUGE edits >15KB easily truncated, write half then append").

## Solution

Make `do_file_write` truncation-aware:

1. **Detect truncation**: unclosed `<file_content>` tag + llmcore truncation 
   markers (`max_tokens !!!]` / `[!!! 流异常中断` / `!!!Error:`) at content tail.
2. **Strip markers**, write the partial chunk immediately to disk (overwrite 
   first chunk, append subsequent chunks).
3. **Stash resume state** in `self.working['_file_write_resume']`: path, 
   round count, and last 80 chars as overlap anchor.
4. **Return a continuation prompt** instructing the LLM to keep outputting 
   from the cutoff (with the 80-char tail shown as anchor).
5. **On the next call**, detect resume state, dedup overlap (try 80→1 chars 
   for largest match), append, and either finish or loop again.
6. **Cap at 8 rounds** (~80KB) to prevent runaway loops; report `partial` 
   status if exceeded.

User experience: large file writes "just work" — no manual splitting needed.

## Tests

`tests/test_file_write_resume.py` (standalone, no pytest dep) covers:
1. Normal complete write (closed tag) — no resume
2. First truncation — partial write + resume state set + correct prompt
3. Continuation with overlap — dedup precise, file completed, state cleared
4. Multi-round truncation (3 rounds) — all dedup correct end-to-end

Run: `python tests/test_file_write_resume.py` → `🎉 ALL 4 TESTS PASSED`

## Scope

- Only touches `ga.py:do_file_write` (+81/-12 lines) and adds one test file.
- No API/schema change; tool description simplified (removes the now-obsolete 
  "HUGE edits easily truncated" warning).
- `file_patch` is unchanged.

### DIFF
--- a/ga.py
+++ b/ga.py
@@ -371,31 +371,100 @@ class GenericAgentHandler(BaseHandler):
         path = self._get_abs_path(args.get("path", ""))
         mode = args.get("mode", "overwrite")  # overwrite/append/prepend
         action_str = {"prepend": "Prepending to", "append": "Appending to"}.get(mode, "Overwriting")
-        yield f"[Action] {action_str} file: {os.path.basename(path)}\n"
 
+        # --- Truncation-aware content extraction ---
         def extract_robust_content(text):
+            """Extract file content, tolerating truncated (unclosed) <file_content> tags."""
+            # 1. Normal closed tag
             tags = re.findall(r"<file_content[^>]*>(.*?)</file_content>", text, re.DOTALL)
-            if tags: return tags[-1].strip()
+            if tags: return tags[-1].strip(), False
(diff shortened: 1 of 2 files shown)


lsdefine:
now it does not use <file_content>

PR #277, opened 2026-05-06
fix: correct GenericAgent typo, improve error handling, expand English docs

GGzili:
Three small, targeted fixes that follow the contributing guidelines:

## 1. Fix class name typo: GeneraticAgent → GenericAgent

The class name was consistently misspelled across 14 files (agentmain.py + 13 frontends). Since the project is called GenericAgent, the internal class should match. Pure rename, zero behavioral change.

## 2. Replace bare except clauses with specific exceptions

- agentmain.py: bare except: pass that silently swallowed session init failures now logs a warning with the key name and error
- agentmain.py: bare except: in next_llm changed to except Exception: to avoid catching KeyboardInterrupt/SystemExit
- ga.py: bare except: in do_code_run changed to except (ValueError, TypeError): since only int conversion can fail there

## 3. Expand English mykey template (77 to ~370 lines)

The Chinese mykey_template.py has comprehensive documentation including session type reference, native vs non-native explanation, apibase rules, complete field reference, and provider-specific examples. The English version was ~77 lines with minimal guidance. This brings parity with the Chinese template so international users can configure GA correctly without needing to read Chinese.

All changes are compact, self-documenting, and have minimal change radius.

### DIFF
--- a/agentmain.py
+++ b/agentmain.py
@@ -39,7 +39,7 @@ def get_system_prompt():
     prompt += get_global_memory()
     return prompt
 
-class GeneraticAgent:
+class GenericAgent:
     def __init__(self):
         os.makedirs(os.path.join(script_dir, 'temp'), exist_ok=True)
         self.lock = threading.Lock()
@@ -55,7 +55,7 @@ class GeneraticAgent:
         mykeys, changed = reload_mykeys()
         if not changed and hasattr(self, 'llmclients'): return
         try: oldhistory = self.llmclient.backend.history
-        except: oldhistory = None
+        except Exception: oldhistory = None
         llm_sessions = []
         for k, cfg in mykeys.items():
             if not any(x in k for x in ['api', 'config', 'cookie']): continue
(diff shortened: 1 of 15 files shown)


lsdefine:
Thanks for the contribution. A few notes on why we're closing this:

1. **Class rename**: Already addressed in main with a backward-compatible alias (`GeneraticAgent = GenericAgent`). Your approach of hard-renaming across all frontends would break external scripts that import the old name.

2. **Exception cleanup**: Valuable, but the scope is too small to justify a multi-file PR. We'll incorporate these fixes directly.

3. **English template expansion**: The direction is right (parity with the Chinese template), but 77→400 lines is too aggressive for a config template. This deserves a separate, focused PR with tighter editing — ideally keeping the template minimal and moving detailed examples to docs.

General note: please avoid bundling unrelated changes in a single PR. Each of these three topics has different review criteria and merge risk.

Appreciate the effort — feel free to re-submit the English docs expansion as a standalone PR.

PR #284, opened 2026-05-07
docs: improve English mykey template guidance

GGzili:
This is a focused follow-up to the previous PR feedback: keep the English mykey template improvement standalone and much tighter.

Changes are limited to mykey_template_en.py:

- Adds a compact session type reference table
- Clarifies Native vs legacy text-protocol sessions
- Adds apibase auto-append rules
- Lists common config fields users need most often
- Keeps examples minimal: Native Claude, Native OAI, mixin, legacy session, proxy, chat integrations, Langfuse

This intentionally avoids the prior broad 77→400 line expansion and does not include class renames or exception cleanup.


lsdefine:

PR #291, opened 2026-05-07
fix: handle queue.Empty in task mode to prevent crash on LLM timeout

Jeason00011:
## Problem

`--task` mode crashes with unhandled `queue.Empty` when LLM response has no chunk output for 300s.

```python
# agentmain.py:211 (before fix)
while 'done' not in (item := dq.get(timeout=300)):   # raises queue.Empty → crash
```

Trigger: `stream=False` + slow model (queuing/rate-limited), or network stall where TCP hasn't timed out yet.

## Evidence: reflect mode already handles this

```python
# agentmain.py:240-245 (reflect mode)
try:
    while 'done' not in (item := dq.get(timeout=120)): pass
    result = item['done']
except Exception as e:
    if getattr(mod, 'ONCE', False): raise
    print(f'[Reflect] drain error: {e}'); result = f'[ERROR] {e}'
```

Same pattern, same timeout mechanism, but wrapped in try/except. Task mode was missed.

## Fix

Add try/except around the task mode drain loop. On timeout, write error to output file and break gracefully (consistent with reflect mode behavior).

## Discovery

Found by systematic code health scan — checking symmetric code paths for inconsistent error handling (hint from repo maintainer to inspect the dual-mode pattern).


lsdefine:

PR #293, opened 2026-05-07
feat: auto-load recent sessions after restart (#218)

### LINKED ISSUE #218: feat: auto-load recent sessions after restart to avoid cold-start memory loss
## Scenario

If GenericAgent is used as a long-running companion-style or IM-entry agent, restarts are normal: system updates, dependency upgrades, model configuration changes, code changes, crash recovery, and so on. If the agent remembers nothing from the last few turns after restart, users will feel that the agent has lost its memory.

## Current Pain Points

- After restart, only long-term memory remains; concrete recent conversations are lost.
- Context, constraints, and task progress that the user just provided cannot be restored automatically.
- Asking users to manually find logs or restate everything leads to a poor experience.
- Injecting all old logs directly into the prompt can also be dangerous, because old state may be mistaken for current instructions.

## Suggested Direction

Add cold-start recent session recall:

- On startup, scan the most recent N conversation logs / model response files.
- For each session, extract only the last few user/assistant turns or an existing summary.
- Clearly mark the content as “recent session memory”, not current instructions.
- Inject it only into the first non-lightweight message to avoid repeated prompt pollution.
- Make it configurable, for example `GA_STARTUP_RECALL_SESSIONS=5` and `GA_STARTUP_RECALL_LINES=8`.

Recent session recall should be separated from long-term memory. It should only help orientation after cold start, and should not automatically be promoted into durable facts. Stable preferences and methodologies should still go into memory/skills.

## Acceptance Criteria

- After restart, the first real message can include a recent-session summary.
- Recent-session content includes timestamp/source and clearly says “do not treat this as a new instruction”.
- Lightweight greetings/status checks do not trigger large history injection.
- The feature can be disabled through an environment variable or config option.

succ985:
## Summary

After restart, automatically inject recent session context into the first user message so the agent doesn't lose conversational memory. Fixes #218.

## Changes

- **`frontends/startup_recall.py`** (new, ~150 lines): Scans `model_responses/` for recent sessions, extracts real user-agent conversations, formats recall context.
- **`agentmain.py`** (+10 lines): Centralized hook in `GenericAgent.run()` — triggers once per restart on first user message only.

## Design Decisions

| Decision | Reason |
|----------|--------|
| Hook in `agentmain.run()` not per-frontend | One code path covers all frontends (wechat/tg/feishu/discord/qq) |
| `source == 'user'` filter | No injection for reflect/scheduler/internal tasks |
| History stays clean | Recall injected into `raw_query` after `history.append()`, so session history isn't polluted |
| Env vars only, no mykey.py | Simpler config: `GA_STARTUP_RECALL_SESSIONS=5`, `GA_STARTUP_RECALL_LINES=8` (set SESSIONS=0 to disable) |
| Regex exact-match for trivial messages | `^(hi|你好|?)\s*$` — won't false-positive on "帮我看看状态" |

## Config

| Env Var | Default | Description |
|---------|---------|-------------|
| `GA_STARTUP_RECALL_SESSIONS` | `5` | Recent sessions to scan. `0` disables feature |
| `GA_STARTUP_RECALL_LINES` | `8` | Conversation turns per session to extract |

## Testing

- ✅ Syntax check: all files pass `py_compile`
- ✅ Unit tests: trivial message detection, disable via SESSIONS=0, recall extraction
- ✅ End-to-end: restarted wechatapp.py 4 times, verified recall injection on first message each time
- ✅ History integrity: `self.history` does not contain recall content

## PR Checklist

- [x] Issue #218 linked
- [x] Self-documenting code, minimal comments
- [x] Change radius: 2 files only (1 new, 1 modified)
- [x] No new dependencies
- [x] Net lines: +222 (new module + hook)


lsdefine:

PR #346, opened 2026-05-12
refactor(langfuse/chatapp):

desmonna:
- Fix 1: _patched_log function signature mismatch
  - Location: plugins-langfuse_tracing.py
  - Problem: monkey-patch function only accepts 2 parameters, but the original function has 3 parameters (including the default parameter log_path)
  - Fix: add log_path=None parameter and pass it to the original function

- Fix 2: Agentmain module import failed
  - Location: frontends-chatapp_common.py
  - Problem: Modules in the frontends directory cannot be imported into the agentmain of the upper directory
  - Fix: Add the parent directory to sys.path at the beginning of the file

### DIFF
--- a/frontends/chatapp_common.py
+++ b/frontends/chatapp_common.py
@@ -1,5 +1,10 @@
 import ast, asyncio, glob, json, os, queue as Q, re, socket, sys, time
 
+# 确保能导入上级目录的模块（如 agentmain）
+_parent_dir = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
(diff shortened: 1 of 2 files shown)


lsdefine:
Thank you for your contribution! This PR modifies **2 files**, but the core files are changed by **209%** and **173%** respectively — effectively complete rewrites. Changes of this magnitude to core files cannot be reviewed safely in a single PR and violate our impact-radius requirements. Please break this into incremental changes and resubmit.

desmonna:
好的，已修改后重新推送

PR #357, opened 2026-05-13
fix: preserve thinking blocks with empty signatures for Mimo API compatibility

C0dethewor1d24:
## Problem
Some API providers (e.g. Xiaomi Mimo) return thinking blocks with empty `signature` fields in their Anthropic-compatible responses. The existing `_drop_unsigned_thinking()` treats empty signatures as missing and strips these blocks from conversation history.

When the API receives subsequent requests without the expected thinking content, it returns HTTP 400: `"The reasoning_content in the thinking mode must be passed back to the API."`

## Fix
Add `_fill_empty_signatures()` to assign a placeholder signature (`"placeholder"`) to thinking blocks that have actual thinking content but an empty signature. These blocks then survive the `_drop_unsigned_thinking()` filter and are correctly passed back to the API.

The fix is a no-op for providers that already return proper signatures (e.g. official Anthropic API).

### DIFF
--- a/llmcore.py
+++ b/llmcore.py
@@ -358,7 +358,7 @@ def _stream_with_retry(sess, url, headers, payload, parse_fn):
     for attempt in range(sess.max_retries + 1):
         streamed = False
         try:
-            with requests.post(url, headers=headers, json=payload, stream=sess.stream, 
+            with requests.post(url, headers=headers, json=payload, stream=sess.stream,
                                timeout=(sess.connect_timeout, sess.read_timeout), proxies=sess.proxies, verify=sess.verify) as r:
                 if r.status_code >= 400:
                     if r.status_code in _RETRYABLE and attempt < sess.max_retries:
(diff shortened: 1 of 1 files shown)


lsdefine:
Thanks for the contribution! Your analysis is spot on — Mimo returns thinking blocks with empty signatures, and `_drop_unsigned_thinking` was incorrectly stripping them, causing 400 errors on the next turn.

However, I've already addressed this with a different approach that I believe is cleaner: instead of patching the signature with a placeholder, I made `_drop_unsigned_thinking` conditional on the model name:

```python
messages = _fix_messages(messages)
if 'claude' in model.lower(): messages = _drop_unsigned_thinking(messages)
messages = _ensure_thinking_blocks(messages, self.model)
```

This fundamentally solves the problem for all non-Claude models (Mimo, DeepSeek, etc.) — `_drop_unsigned_thinking` only exists to satisfy Anthropic's signature verification requirement, so it should only run when actually talking to Claude's API. This approach is semantically clearer and has no side effects (a placeholder signature would fail validation if ever sent to the real Claude API).

Closing this PR since the fix is already in place. Thanks again for identifying the issue and taking the time to submit a fix!

PR #360, opened 2026-05-13
feat(tui): v2 image paste folding, history nav, /llm & /export choice…

shenhao-stu:
… list

Several small UX improvements for the Textual v2 frontend, all isolated to `frontends/tuiapp_v2.py`. No core agent / IO contract changed; images are still passed to the model as plain path text.

Image paste / clipboard
- Ctrl+V (or drag-drop) on the input now saves the clipboard bitmap to `tmp/<ts>.png` and inserts a folded placeholder `[Image #N]`. The real path is kept in a per-app `_pastes` map and expanded only at submit time, so users can mouse-select & Ctrl+C the placeholder without leaking the OS path into the terminal selection buffer.
- `_IMAGE_RE` still matches the long form `[Image #N: <path>]` for backward compatibility with older pasted history.

Interactive command palettes (ccstatusline-style)
- `/llm` with no args now shows an inline ChoiceList of channels; Up/Down to move, Right/Enter to confirm, Left/Esc to cancel.
- `/export` with no args shows three choices: clip / all / file (auto timestamp). Explicit forms are preserved and extended:
    /export clip|copy        copy last reply wrapped in code fences
    /export all              print full log file path
    /export file [name]      export last reply; name optional
    /export <name>           legacy short form, still works
- Both palettes collapse to a single-line "已选 …" / "（已取消）" row
  so the scroll buffer stays clean.

Input history navigation
- Up/Down in the input cycles through previous submissions, mirroring bash/zsh. The built-in command palette popup is suppressed while navigating via `_suppress_palette_open`, so the two no longer fight for the same keys.

Misc
- Ctrl+U clears the current input (parity with readline).
- All new keybindings use Textual's portable Binding API, no platform-specific code, works on Win/macOS/Linux terminals.

### DIFF
--- a/frontends/tuiapp_v2.py
+++ b/frontends/tuiapp_v2.py
@@ -17,6 +17,7 @@ import os
 import queue
 import re
 import sys
+import tempfile
 import threading
 import time
 from dataclasses import dataclass, field
@@ -139,6 +140,7 @@ class ChatMessage:
     choices: list = field(default_factory=list)   # [(label, value), ...]
     on_select: Optional[Callable] = field(default=None, repr=False)
     selected_label: Optional[str] = None
+    image_paths: list[str] = field(default_factory=list)
     # Mounted widget refs (None until mounted)
     _role_widget: Any = field(default=None, repr=False)
     _hint_widget: Any = field(default=None, repr=False)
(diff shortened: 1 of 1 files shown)


lsdefine (approved):
✅ Approved via agent review. Changes clean, self-contained in tuiapp_v2.py (+320/-41), no core coupling. LGTM.

PR #403, opened 2026-05-16
feat(tui v2): theme system + Markdown/streaming/continue fixes

nianyucatfish:
## Summary

  Adds a theme system to TUI v2 and fixes a cluster of rendering issues that surfaced along the way.

  **Themes**
  - Themable color system wired through Textual's theme registry; `ga-default` (github-dark) plus nord / gruvbox / dracula / tokyo-night / textual-light.
  - `Ctrl+T` opens a live-preview theme picker; choice persists to `temp/tui_settings.json` and is restored on startup.
  - Non-`ga-default` palettes use a minimal Markdown theme (fg/muted only) so body text stays readable until each theme is hand-tuned.
  - Theme switch is cache-only invalidation (no full remount), so switching is instant.

  **Markdown / rendering**
  - Anchor `_META_TAG_RE` to line start so `<summary>` / `<thinking>` tags inside backticks survive.
  - CJK-aware wrap patched on every Rich + Textual binding (Rich Markdown, user/system messages, input box).
  - Clean copy: parallel wide render builds a wrap-free source so selection→copy never includes visual wrap newlines.

  **Streaming perf**
  - Defer Markdown render until streaming completes — per-chunk path updates plain `Text`; the terminal `done` chunk renders Markdown once and swaps.
  Drops O(chunks × turn_len × 2) Markdown parses to O(turn_len × 2).

  **/continue restore**
  - Rebuild restored transcripts to match live `agent_loop` output (tool_use headers + tool_result fences).
  - Classify auto-continuation prompts by `tool_result` presence so they don't show up as fake user messages.

  **Other small fixes**
  - Sidebar Q/S preview switched to `C_MUTED` so it stays readable on the active row under tokyo-night.
  - Drop the `theme: <name>` system toast on every theme change.
  - Align `ga-default` chip_time with the sidebar's active-session green.
  - Hook topbar chip colors into the theme and move the fold indicator to the right column.

### DIFF
--- a/frontends/continue_cmd.py
+++ b/frontends/continue_cmd.py
@@ -246,7 +246,9 @@ def _user_text(prompt_body):
 
 
 def _assistant_text(response_body):
-    """Joined text from a response blocks repr; '' on parse failure."""
+    """Joined plain text from a response blocks repr; '' on parse failure.
+    Used by /export to grab the model's prose only, without tool noise.
+    """
     try: blocks = ast.literal_eval(response_body)
     except Exception: return ''
     if not isinstance(blocks, list): return ''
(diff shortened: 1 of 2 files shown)


lsdefine:
LGTM. 主题系统和streaming优化都很实用，merge了。

PR #411, opened 2026-05-17
feat(tui v2): plan card + done annotation + /cost + /review wiring

shenhao-stu:
- Plan card (📋): per-session message-scan with `plan_scan_baseline` so `/continue` of a finished session doesn't auto-resurrect old plan cards; prefers `<summary>当前步骤:` over the long plan-item echo; placeholder while enter_plan_mode is armed but plan.md hasn't been written yet.
- Done annotation (⠿ {Verb} for Xm Ys · ↑ N · ↓ M): post-turn elapsed
  + per-call token deltas, frozen at done so the line stays put across re-mounts. Spinner annotation gated on cumulative deltas past stream-start baselines — fixes `↓ 16` showing immediately on prompt submit.
- fold_turns regex anchored to line-start (`^` + MULTILINE) so embedded backticks in tool output (e.g. `N|\`\`\`\`` gutter from file_read) don't pair with later real fences. ~100x faster on long /continue sessions.
- /cost: per-session card + `/cost all`; `Context window` line now reads GA's actual char cap via `cost_tracker.context_window_chars(backend)` = `backend.context_win * 3` (the `trim_messages_history` trigger threshold in llmcore.py), so the percentage reflects real budget and respects `mykey.py` overrides. Unit changes from token to character. Dropped the model-theoretical `_CTX_LIMITS` table (never actually constrained anything). Subagent log scan via `cost_tracker.scan_subagent_logs` aggregates `temp/*/stdout.log` since TUI start.
- /review: standalone in-session reviewer; reuses `review_cmd.handle` to render the inline prompt and submits as a normal user task.
- llmcore: emit `[Output] tokens=N` for chat_completions / responses modes (already emitted for `messages` SSE) so cost_tracker scopes output tokens uniformly across all three API modes.
- review_inline_prompt.{txt,en.txt}: filled out the in-session review discipline — Three Iron Rules, Q1-Q4 framing, severity examples, 8 false-positive rules expanded, 8 wording rules expanded, 8-item self-check. ZH/EN both grow ~80 lines.
- memory/review_sop.md: rewrite 52 → 191 lines, 11 sections covering quick-start, entry files, Three Iron Rules, 5-step workflow, severity, output protocol, relationship to plan_sop / verify_sop. Style mirrors plan_sop.md + autonomous_operation_sop.md.

Files: frontends/{tuiapp_v2,tuiapp,cost_tracker,plan_state}.py, llmcore.py, memory/review_sop.md, memory/review_sop/review_inline_prompt.{txt,en.txt}, .gitignore (allowlist new SOP).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -101,6 +101,9 @@ memory/L4_raw_sessions/*
 # Code Review Principles
 !memory/code_review_principles.md
 
+# Review Mode SOP
+!memory/review_sop.md
+
 # Visual Studio
 .vs/
 restore_commit.txt
--- a/frontends/cost_tracker.py
+++ b/frontends/cost_tracker.py
@@ -1,13 +1,14 @@
-"""Per-thread LLM token usage, captured via llmcore monkey-patches.
+"""Per-thread LLM token usage via llmcore monkey-patches.
 
-`install()` wraps `llmcore._record_usage` (covers all three API modes) and
-`llmcore.print` (the `messages` SSE path emits the final `output_tokens`
-only via `[Output] tokens=N`, never through `_record_usage`). Tracking is
-keyed by `threading.current_thread().name`; each TUI session runs the
-agent on a uniquely named thread (`ga-tui-agent-<id>`), so `/cost` is a
-thread lookup.
+`install()` wraps `llmcore._record_usage` + `llmcore.print` (the SSE
+`messages` path only emits final `output_tokens` through `[Output] tokens=N`).
+Trackers are keyed by `threading.current_thread().name`; each TUI session
+runs its agent on `ga-tui-agent-<id>`, so `/cost` is a thread lookup.
+
+Subagent processes are out-of-process, so `scan_subagent_logs` parses the
+same `[Cache]` / `[Output]` print lines from `temp/*/stdout.log`.
 """
-import re, threading, time
+import glob, os, re, threading, time
 from dataclasses import dataclass, field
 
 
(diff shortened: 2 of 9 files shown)


lsdefine:
Thanks for the PR. I checked the diff and noticed that this PR currently contains changes beyond the TUI v2 plan/cost/review feature scope.

Please avoid mixing unrelated changes into this PR. In particular, the `tmwd_cdp_bridge` changes appear unrelated to the plan card / `/cost` / `/review` work and should be moved to a separate PR if still needed. Keeping this PR focused will make the review safer and much easier to reason about.

Also, please consider organizing the new frontend helper modules under a dedicated subfolder inside `frontends/`. Right now files that are indirectly called by `chatapp_common` / TUI code, such as review/cost/plan helpers, are placed directly under `frontends/`. As this grows, it would be cleaner to put these second-level helper modules into a subpackage, for example:

```text
frontends/
  common/
    plan_state.py
    cost_tracker.py
    review_cmd.py
```

or another name that matches the intended boundary. This would make the module ownership clearer and avoid further cluttering the top-level `frontends/` directory.

Also, please double-check the `/cost` accounting path. I noticed a possible double-counting risk around output tokens:

- `llmcore._record_usage()` now emits `[Output] tokens=N` for `responses` and `chat_completions`.
- `cost_tracker.install()` monkey-patches `llmcore.print` and increments output when it sees `[Output] tokens=N`.
- If `cost_tracker.record_patched()` also increments output directly for the same `_record_usage()` call, output tokens may be counted twice.

Maybe the current implementation already avoids this for some modes, but the interaction is subtle enough that it needs either a small regression test or a clearer invariant in the code. Suggested test cases:

1. `responses` usage with `output_tokens=N` should increase output by exactly `N`, not `2N`.
2. `chat_completions` usage with `completion_tokens=N` should increase output by exactly `N`, not `2N`.
3. `messages` streaming path should still count the final `[Output] tokens=N` exactly once.

So my requested changes are:

1. Split unrelated `tmwd_cdp_bridge` changes out of this PR.
2. Move the second-level frontend helper modules into a dedicated `frontends/` subfolder/subpackage.
3. Add/fix a regression test for `/cost` output-token accounting to ensure no double counting.

PR #459, opened 2026-05-22
Add plan-mode pending guard to enforce enter_plan_mode after reading plan_sop

### LINKED ISSUE #458: Plan mode can be bypassed after reading plan_sop without enter_plan_mode
## Problem

When an agent reads `memory/plan_sop.md` (the standard operating procedure for plan mode), it is supposed to subsequently create a plan file and enter plan mode by calling `handler.enter_plan_mode(...)`. In practice, the agent may skip this step and continue normal execution. The system does not enforce the transition—plan mode is effectively opt-in with no runtime guard.

## Root cause

The existing implementation only enforces plan-mode constraints *after* `enter_plan_mode()` has been called:

- `do_no_tool()` blocks premature completion claims only when already in plan mode.
- `turn_end_callback()` appends the plan hint only when `_in_plan_mode()` is true.

There is no mechanism to detect that plan mode *should have been entered but was not*, so the agent can read `plan_sop`, extract the SOP points, and then proceed with ordinary tool calls—without ever creating `plan.md` or calling `enter_plan_mode()`.

## Impact

Without a guard, complex tasks lack the mandatory plan→execute→verify lifecycle:

- No `plan.md` is created or maintained.
- No per-step checklist enforcement.
- The verify/verdict gate is bypassed.
- Existing plan-mode UI and completion checks are never activated.

## Proposed fix

Add a pending-state guard between "plan SOP read" and "plan mode entered":

1. When the agent reads `plan_sop` (or otherwise triggers a plan-mode-required task), set a pending flag.
2. If the next turn still has not called `handler.enter_plan_mode(...)`, the turn-end callback injects a hard guard: create plan, enter plan mode, and disallow ordinary execution until done.
3. Once `enter_plan_mode()` succeeds, clear the pending state.
4. All existing in-plan-mode checks continue to work as before.

## Acceptance criteria

- After reading `plan_sop`, if `enter_plan_mode()` is not called, the next turn is intercepted and corrected.
- `self.working["in_plan_mode"]` is set to the plan file path.
- TUI plan/todo panel recognizes the active plan.
- Unchecked items prevent premature completion.
- Tests cover: pending flag set on SOP read, guard intercepts missing entry, pending cleared on successful entry.

HamsteRider-m:
## What

When `plan_sop.md` is read, a pending plan-mode flag is now set. If the agent does not call `handler.enter_plan_mode(...)` in the same turn, the turn-end callback injects a hard guard that blocks normal execution until plan mode is properly entered. Once `enter_plan_mode()` succeeds, the pending state is cleared and all existing in-plan-mode checks take over.

This prevents the agent from reading the SOP, extracting its points, and then continuing ordinary execution without ever creating a plan file or entering plan mode.

## Changes

- `ga.py`: set `plan_mode_pending` on plan SOP read; inject guard in `turn_end_callback()` when pending but not yet in plan mode; clear pending on `enter_plan_mode()`.
- `test_plan_mode_guard.py`: three test cases covering pending flag set, guard interception, and pending cleared on entry.

## Verification

```
python -m pytest -q test_plan_mode_guard.py   # 3 passed
python -m py_compile ga.py test_plan_mode_guard.py   # ok
```

Closes #458


lsdefine:

PR #478, opened 2026-05-25
fix(tgapp): support /review command handling

YooooEX:
## Context

Telegram `/help` and the bot command menu already expose `/review`, but Telegram command messages are consumed by `handle_command()` before they can reach the shared slash-command hook. This PR wires `/review` into the Telegram command handler while keeping the change radius inside `frontends/tgapp.py`. No new dependencies are added.

## Summary

Add Telegram `/review` slash-command handling by reusing the shared `review_cmd` prompt path and streaming generated review prompts through the existing Telegram response flow.

### `frontends/tgapp.py`

- Import the shared `/review` handler from `review_cmd`.
- Add Telegram-local `/review` argument parsing so `/review` and `/review <scope>` map to the shared review command body.
- Route `/review help` style responses through the existing MarkdownV2-safe command reply path.
- For review prompts, cancel any active Telegram stream task before starting the review task through `agent.put_task(..., source="telegram")`.
- Reuse the existing `_stream(...)` lifecycle and `ctx.user_data["stream_task"]` tracking for review output.


lsdefine (approved):
Approved: wires existing /review command support into Telegram frontend; implementation matches review_cmd.handle contract and compiles.

PR #483, opened 2026-05-25
feat: prefer pwsh (PowerShell 7) over powershell in code_run on Windows

desmonna:
Use shutil.which('pwsh') to detect PowerShell 7 availability, falling back to legacy powershell if not found. PS7 has better UTF-8 support, modern syntax, and is the recommended shell for Windows development.

### DIFF
--- a/ga.py
+++ b/ga.py
@@ -1,7 +1,7 @@
 import sys, os, re, json, time, threading, importlib
 from datetime import datetime
 from pathlib import Path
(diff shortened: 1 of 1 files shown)


desmonna:
Fixes #484

lsdefine (approved):
Approved: uses shutil.which to prefer pwsh when available while safely falling back to legacy powershell.

PR #486, opened 2026-05-25
fix: wrap _turn_end_hooks iteration with list() to prevent RuntimeError

w286554018:
When a hook modifies `_turn_end_hooks` dict during iteration in `turn_end_callback`, Python raises `RuntimeError: dictionary changed size during iteration`.

This fix wraps the `.values()` call with `list()` to create a snapshot of the hooks before iterating, preventing the RuntimeError.

**Changes:**
- `ga.py` line 576: `for hook in getattr(self.parent, '_turn_end_hooks', {}).values()` → `for hook in list(getattr(self.parent, '_turn_end_hooks', {}).values())`


lsdefine (approved):
Approved: small safe fix to snapshot hook values during iteration.

Issue #351, opened 2026-05-12
tuiapp_v2.py: Copy and paste do not work in the TUI


liuyuyan6100:
## Description

I found that the new Textual-based TUI interface `tuiapp_v2.py` does not support normal copy and paste behavior in local testing.

This affects basic terminal interaction, especially when using the TUI for long prompts, multi-line command input, or copying generated output.

## Affected File

- `frontends/tuiapp_v2.py`

## Steps to Reproduce

1. Pull the latest code:

   ```bash
   git pull
   ```

2. Install Textual if needed:

   ```bash
   pip install textual
   ```

3. Run the new TUI from the project root:

   ```bash
   python frontends/tuiapp_v2.py
   ```

4. Try the following operations:

   - Paste text into the input area
   - Paste multi-line text into the input area
   - Select text from the TUI output area and copy it
   - Use common terminal copy/paste shortcuts

## Expected Behavior

Copy and paste should work normally:

- Users should be able to paste single-line text into the input box
- Users should be able to paste multi-line text into the input box
- Users should be able to select and copy text from the TUI output area
- Existing TUI shortcuts should not block basic terminal copy/paste behavior

## Actual Behavior

Copy and paste do not work as expected in `tuiapp_v2.py`.

In my testing, basic copy/paste behavior appears to be blocked or unavailable in the new TUI interface.

## Impact

This significantly affects the usability of the new TUI, especially for:

- Long prompt input
- Multi-line text editing
- Copying generated responses
- Reusing command outputs
- Daily terminal-based workflows

## Environment

- OS: Ubuntu 24.04.4 LTS
- Python: Python 3.12.3
- Textual: 8.2.5
- Shell: `/bin/bash`
- TERM: `xterm`
- SSH: yes
- tmux: no
- Local commit tested: `5b50eac`

## Additional Notes

The previous `tuiapp.py` does not appear to be affected in the same way.

Since `tuiapp_v2.py` was recently introduced as the new Textual-based interface, it would be great to ensure that copy/paste remains compatible with common terminal workflows.

I am reporting this after local verification and would be happy to help test a fix or submit a PR if the expected behavior is confirmed.

szdtzpj:
Submitted draft PR #354 with a focused fix for this: https://github.com/lsdefine/GenericAgent/pull/354

The change restores `Ctrl+V` paste in `tuiapp_v2.py` while keeping bracketed-paste newline normalization, long multi-line paste placeholders, and duplicate-paste protection.

[...]

PR #491, opened 2026-05-26
fix: use OSC52 clipboard for remote macOS TUI

HamsteRider-m:
## Summary
- keep the existing macOS `pbcopy` path for local Terminal.app sessions
- fall back to Textual/OSC52 when the TUI appears to be running through SSH or tmux
- add a `GA_TUI_CLIPBOARD=osc52` escape hatch for users who need to force terminal clipboard output

## Why
`frontends/tuiapp_v2.py` currently returns immediately after `pbcopy` succeeds on macOS. When the TUI runs on a remote Mac over SSH/tmux, that copies to the remote Mac clipboard instead of the user's visible terminal clipboard. Allowing OSC52 in remote/tmux sessions lets compatible terminals/tmux propagate the copy to the local clipboard.

Refs #351.

## Testing
- `PYTHONPYCACHEPREFIX=/tmp/ga-pr-pycache python -m py_compile frontends/tuiapp_v2.py`
- `python frontends/tuiapp_v2.py --help`
- `git diff --check`
- monkeypatch smoke test for local macOS, SSH, tmux, and `GA_TUI_CLIPBOARD=osc52` paths

### DIFF
--- a/frontends/tuiapp_v2.py
+++ b/frontends/tuiapp_v2.py
@@ -2170,17 +2170,18 @@ class GenericAgentTUI(App[None]):
         self._refresh_all()
 
     def copy_to_clipboard(self, text: str) -> None:
-        """Override Textual's OSC 52 clipboard (broken on macOS Terminal.app).
-        Use pbcopy on macOS, fall back to OSC 52 on other platforms."""
-        import sys, subprocess as _sp
-        if sys.platform == "darwin":
+        """Copy locally on macOS, and use OSC 52 when the UI may be remote."""
+        import subprocess as _sp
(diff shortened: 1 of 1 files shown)


shenhao-stu:
这个问题已经被 #418 修过了，commit 54f40d9 `fix(tui): use pbcopy on macOS for clipboard (OSC 52 broken on Terminal.app)`。

当前实现：`frontends/tuiapp_v2.py` 的 `copy_to_clipboard()` —— macOS 走 pbcopy，其他平台 OSC52 fallback，与你提的方向一致。先关掉，感谢～

PR #493, opened 2026-05-26
 fix(llmcore): add Cloudflare 520-527 to _stream_with_retry _RETRYABLE set

shenhao-stu:


lsdefine (approved):
Approved via API after local merge/py_compile checks. Merge order: #493 -> #495 -> #496.

PR #497, opened 2026-05-27
fix(tui): preserve selection after Ctrl+C copy

yumengv:
## Summary
- Preserve the visible selection after Ctrl+C copies selected text in TUI v3.
- Keep the original no-selection behavior unchanged: first Ctrl+C clears the draft and arms quit; second Ctrl+C exits.

## Testing
- `python -m py_compile frontends/tui_v3.py`
- Static assertion: selection-copy branch no longer assigns `self._sel = None`.
- Static assertion: no-selection quick-clear branch still exists.


lsdefine:

PR #507, opened 2026-05-27
fix(tui): slash-cmd polish, bundle i18n sync & settle the spinner on abort

shenhao-stu:
## Summary

Three independent TUI fixes that surfaced while polishing the slash-command experience. No behavioural change to the agent core — all edits live under `frontends/`.

- **`/scheduler` running-marks perf** — stop re-scanning every process's full cmdline on each open.
- **v3 bundle-command i18n** — `/update /autorun /morphling /goal /hive /scheduler` rows now translate instead of showing hardcoded English.
- **Abort/stop spinner** — Ctrl+C / `/stop` now *settles* the live spinner instead of letting it keep ticking elapsed while the LLM stream unwinds.

2 commits · 3 files · +134 / −44.

---

### `frontends/slash_cmds.py` — `/scheduler` running-marks perf
- Add a **2s TTL snapshot** (`_RUNNING_CACHE`) + `invalidate_running_cache()`.
- `running_services(use_cache=True)`: drop `"cmdline"` from the bulk `process_iter` fetch and only call `proc.cmdline()` **after** the python-ish name pre-filter — the per-proc cmdline read was the hot cost.
- `start_service` / `stop_service` invalidate the cache so the next read is fresh.
- Net effect: cold ~2.1s → ~1.0s, warm reads ~0ms. Still degrades gracefully to `{}` (no running marks) when `psutil` is missing.

### `frontends/tui_v3.py` — bundle-command i18n sync
- Add en + zh `_I18N` keys (`arg` + `desc`) for `/update`, `/autorun`, `/morphling`, `/goal`, `/hive`, and `/scheduler`.
- `_cmds()` bundle rows now route through `_t()` instead of inline English, so the zh and en command palettes stay in sync (v3 had only English here while v2 was already localized).

### `frontends/tuiapp_v2.py` — settle the spinner on abort (+ DRY + minor)
**Bug:** aborting a turn (Ctrl+C / `/stop`) left the live spinner climbing elapsed seconds until the LLM stream finally unwound in the background.

- New `ChatMessage._stop_summary` — a frozen `(elapsed, in, out)` snapshot stamped the instant the user aborts.
- `_cmd_stop` calls the new `_mark_stopping(sess)`, which freezes **every** in-flight assistant and repaints it to a settled `⠿ Stopping… (Xm Ys · …)` line right away.
- `_spinner_annotation` short-circuits to `_stopping_annotation` once aborted, so elapsed stops ticking immediately.
- `_done_annotation` renders `⠿ Stopped after Xm Ys` (dim) for an aborted turn vs `⠿ {Verb} for Xm Ys` (green) for a normal finish.

**DRY refactor (no behaviour change):**
- `_fmt_tokens()` — single source for the `↑ N · ↓ M` glyph string (previously duplicated across spinner / done / stopping).
- `_freeze_summary()` — single source for the `(elapsed, in, out)` snapshot tuple; `_capture_done_summary` / `_capture_stop_summary` now just delegate.

**Minor (carried on this branch):**
- `ask_user` multi-choice: collapse the picked list to a string (`"; ".join(v)`) before handing it to the agent, matching the breadcrumb.
- i18n typo: tidy the `/reload-keys` description spacing.

---

## Testing
- `python -m py_compile frontends/tuiapp_v2.py` — OK.
- Manual: `/stop` mid-stream now shows `Stopping…` instantly and resolves to `Stopped after …`; normal turns still read `{Verb} for …`.
- `/scheduler` running marks render correctly and noticeably faster on open.

### DIFF
--- a/frontends/slash_cmds.py
+++ b/frontends/slash_cmds.py
@@ -306,6 +306,7 @@ def start_service(name: str) -> tuple[bool, str]:
         rc = proc.poll()
         if rc is not None:
             return False, f"启动失败 (退出码 {rc}): {svc['name']}"
+        invalidate_running_cache()
         return True, f"已启动 {svc['name']} (pid={proc.pid})"
     except Exception as e:
         return False, f"启动失败: {type(e).__name__}: {e}"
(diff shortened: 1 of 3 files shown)


lsdefine (approved):
LGTM - TUI-only changes, good to merge.

PR #508, opened 2026-05-27
fix(tui_v2): /continue resumes unfinished plan + /stop refills input

nianyucatfish:
/continue previously pushed plan_scan_baseline past every restored message, so plan_state.is_active() saw an empty slice and hid the card — even when the prior session's plan.md still had `[ ]` items the user wanted to keep working on. `working['in_plan_mode']` isn't replayed by the log restore either, so the card never came back.

Make the baseline conditional: default to 0 so the scanner walks the full restored history; only push past it when the restored plan.md is already all-done, suppressing the stale ✓ card. Resolved-path missing or empty items naturally fall back to "no card".

/stop now also refills the input box with the just-sent user message, so an interrupted send is one keystroke from edit-and-resend. Only when the box is empty (don't clobber a half-typed follow-up). Agent history is untouched — a resend duplicates the turn in LLM context; /rewind 1 is the manual escape.

Behaviour matrix (/continue):
  unfinished plan  → baseline=0, scanner finds path, card resumes
  completed plan   → baseline=len(msgs), card stays hidden
  plan.md deleted  → resolve_path None, card stays hidden
  fresh enter_plan_mode after continue → working flag re-arms, normal

### DIFF
--- a/frontends/tuiapp_v2.py
+++ b/frontends/tuiapp_v2.py
@@ -3569,6 +3569,8 @@ class GenericAgentTUI(App[None]):
 
     def _cmd_stop(self, args, raw):
         sess = self.current
+        last_user_text = next((m.content for m in reversed(sess.messages)
+                               if m.role == "user"), None)
         try:
             sess.agent.abort()
             if sess.status == "running":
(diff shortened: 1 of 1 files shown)


lsdefine (approved):
LGTM - clean fix for /continue baseline + /stop input refill.

PR #510, opened 2026-05-27
fix(tui): polish tui_v3 Windows UX — 6 issues

shenhao-stu:
1. _TILE_U: switch tile bg/fg to truecolor (48;2;55;55;55 / 38;2;230;230;230) so the file-tree tiles render legibly on Win Terminal dark themes instead of a muddy 8-color block.
2. _ticker: tighten the redraw cadence 0.4s -> 0.1s so the spinner and status line feel responsive on Windows.
3. terminal title: add _set_term_title() helper and emit OSC 0 on init / session new / rename / quit so the tab title tracks the active session.
4. _drain: stream-aware grace window — when the pending stream carries a pending ask_user ("Waiting for your answer"), extend the post-done poll so the question card is not swallowed before the prompt appears.
5. /emoji command: split _PETS into _PETS_UNICODE / _PETS_ASCII behind a module-level _pet_style toggle; add the /emoji dispatch + _cmd_emoji handler so users on fonts without emoji can fall back to ASCII pets.

### DIFF
--- a/frontends/tui_v3.py
+++ b/frontends/tui_v3.py
@@ -1334,9 +1334,12 @@ else:
     _DIM = '\x1b[2m'
     _ACCENT = '\x1b[38;2;94;106;210m'    # Linear lavender #5e6ad2
     _BORDER = '\x1b[38;5;146m'
-_INK_U = '\x1b[38;5;234m'                # user ink — near-black, strong (as requested)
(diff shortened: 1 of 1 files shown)


lsdefine (approved):
TUI-only Windows UX polish. LGTM.

PR #512, opened 2026-05-27
fix(tui_v2): render markdown while streaming

MuziIsabel:
## Summary
- Render assistant Markdown during streaming with stable-prefix caching and bounded tail parsing.
- Temporarily close open fenced code blocks so live code output is styled before completion.
- Keep completed responses on the existing full Markdown render path.

## Test plan
- `python -m py_compile e:\Work\Python\GenericAgent-live-md-pr\frontends\tuiapp_v2.py`
- Helper smoke test for heading split, open fence provisional closure, and table tail split.


lsdefine:

PR #518, opened 2026-05-28
fix(tui): pending-queue inject lands mid-turn (3 s hard cap)

shenhao-stu:
## Summary

User types follow-up lines while the agent is running; queued lines never reach the model — the agent finishes its current run and the queued text spawns a fresh task instead of being spliced in. Two compounding causes:

1. **Cooldown was 5 s.** Longer than many turns. By drain time the agent was already idle, so `_inject_intervene` returned False; the fallback called `put_task`, which starts a new turn rather than injecting into the current one.
2. **Deadline reset on every keystroke.** A steady typer could push the drain off indefinitely. Once they paused, the agent had usually finished.

## Fix

Hard 3 s deadline anchored to the **first** queued entry. Subsequent submits within the window append to the queue without moving the timer. Drain now fires reliably while the current run is still in flight, GA's `turn_end_callback` (ga.py:576) consumes `_intervene` at the next turn boundary, and `[MASTER] ...` is prepended to the upcoming prompt — exactly what mainstream TUIs (Codex CLI Enter-queue, Claude Code) do.

Symmetric change in `tuiapp_v2.py` and `tui_v3.py`.

## Reference
- Codex CLI: Enter queues + drains at next tool call; ESC interrupts and submits immediately (openai/codex#15842, #17095)
- Claude Code: Enter queues; Ctrl+C drains (anthropics/claude-code#44851, #36326)

## Test plan
- [x] `python -c "import ast; ast.parse(open('frontends/tui_v3.py').read())"` syntax OK
- [x] `python -c "import ast; ast.parse(open('frontends/tuiapp_v2.py').read())"` syntax OK
- [x] Type 3 lines in quick succession while a task is running → deadline locks at line-1 + 3 s, all three drain together via `_intervene` before the turn ends.
- [x] Esc on empty input still clears the queue and resets `_pending_drain_t` to 0.
- [x] Up still pops the most-recent queued line back into the composer; deadline clears only when queue becomes empty.

### DIFF
--- a/frontends/tui_v3.py
+++ b/frontends/tui_v3.py
@@ -4784,11 +4784,6 @@ class SB:
                 elif isinstance(ev, (SystemEvent, ErrorEvent)):
                     self._finalize(getattr(ev, 'text', None) or getattr(ev, 'message', '')); break
         self._running = False
-        # Note: pending-drain timer is armed on each user submit (resets
-        # every time they add another), not on turn end — that matches
-        # the "5 s cooldown since LAST keystroke" intent.  Maintenance
-        # loop fires _drain_pending when the deadline lapses regardless
-        # of running state.
         # _ticker stops as soon as _running goes False so the title would
         # freeze on the last spinner frame.  Repaint it now to drop the
         # glyph and reveal a clean idle title.
--- a/frontends/tui_v3.py
+++ b/frontends/tui_v3.py
@@ -2256,16 +2256,13 @@ class SB:
         self._quit = False
         self._cc_t = 0.0                # last bare-Ctrl+C time (arm-to-quit window)
         self._last_esc_t = 0.0          # last bare-Esc time (Esc Esc → /clear)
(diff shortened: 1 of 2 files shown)


shenhao-stu:
Superseded by #521.

#518's 3 s hard cap was the right direction (deterministic deadline anchored to first queue entry) but couldn't solve the root cause demonstrated in `temp/model_responses/model_responses_042511.txt`: ANY mid-turn `[MASTER] ...` prepend to `next_prompt` causes the model to drop the original task and chase the queued message.

Codex CLI (`maybe_send_next_queued_input`, codex-rs/tui/src/chatwidget.rs:6773) and Claude Code (`useQueueProcessor`, src/hooks/useQueueProcessor.ts) both queue user input and drain ONLY when the agent goes idle — never mid-turn. #521 adopts that pattern in both v2 and v3, plus the cross-platform `!cmd` shell fix and v2 right-click cursor fix.

Closing this in favour of #521 to keep review queue clean.

PR #543, opened 2026-05-31
fix(tui): resolve undefined refresh_chrome in _on_stream exit-boundary replay

Kailigithub:
## Problem

In `_on_stream()`, the exit-boundary replay code path (lines 4770-4778) references `refresh_chrome` which is not a parameter of the method. This causes a `NameError` at runtime when:

1. An exit-boundary replay starts a follow-up task before the original display queue emits its final `done`
2. The old done event triggers the stale-task-id branch
3. `if refresh_chrome:` evaluates an undefined name → `NameError`

The issue was introduced during the exit-boundary replay refactoring where `refresh_chrome` was referenced but the variable was only available as a parameter in `_update_assistant()`, not in `_on_stream()`.

## Fix

Since this code path already guards with `if found and agent_id == self.current_id:`, the sidebar/topbar refresh is always appropriate (we're updating the display for the active agent). Removed the undefined `refresh_chrome` conditional and made the chrome refresh unconditional within the `current_id` guard block.

Also fixed the indentation: the display update block was at the wrong level, causing the `refresh_chrome` reference to execute outside the `agent_id == self.current_id` guard.

## Verification

- `ruff check frontends/tuiapp_v2.py --select F821` → All checks passed
- `python3 -m py_compile frontends/tuiapp_v2.py` → No syntax errors

### DIFF
--- a/frontends/tuiapp_v2.py
+++ b/frontends/tuiapp_v2.py
@@ -6125,13 +6125,12 @@ class GenericAgentTUI(App[None]):
                         m.done = True
                         found = m
                         break
-                if found and agent_id == self.current_id:
-                    if found._segment_widgets:
-                        try: self._stream_update_assistant(found)
-                        except Exception: self._refresh_messages()
-                    else:
-                        self._refresh_messages()
-                    if refresh_chrome:
+                    if found and agent_id == self.current_id:
+                        if found._segment_widgets:
+                            try: self._stream_update_assistant(found)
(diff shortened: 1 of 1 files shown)


Kailigithub:
Closing this PR — it's been open for 49 days with no review activity, and I don't want to leave stale work in the queue. Happy to reopen if the change is still wanted; otherwise the diff is preserved at the branch tip and can be cherry-picked at any time.

PR #552, opened 2026-06-01
fix(tui-v2):  /continue命令导致丢失新会话记录

AiHyo:
## Context

In `tui v2`, a newly started conversation disappears from recoverable history after switching to another session with `/continue`.

The active conversation log is stored on `agent.log_path`, but `reset_conversation()` was snapshotting the PID-based fallback path instead. As a result, when `/continue` reset the current conversation before restoring the selected session, the newly started conversation was not preserved.

## Fix

Use the agent's active `log_path` when snapshotting the current conversation before reset, while keeping the existing PID-based path as the fallback for older callers.

## Validation

Completed local testing with `tui v2`: started a new conversation A, switched to conversation B with `/continue`, then used `/continue` again and confirmed conversation A remained recoverable.


lsdefine:

PR #553, opened 2026-06-01
Filter FILE placeholder hints

InkiChang:
What
- Ignore instructional FILE placeholders such as [FILE:filepath] when extracting attachments.
- Deduplicate extracted FILE paths in the shared chat frontend helper.
- Reuse the shared extractor in the Feishu frontend.

Why
- The FILE hint shown to agents can be echoed back literally and was being treated as a missing attachment path in Feishu.
- Filtering placeholder examples prevents frontends that use the shared extractor from trying to send non-existent files.

Testing
- python3 -m unittest tests.test_chatapp_common
- python3 -m py_compile frontends/chatapp_common.py frontends/fsapp.py tests/test_chatapp_common.py


lsdefine:

PR #570, opened 2026-06-05
feat(qq): media in/out, live per-turn heartbeat, and attachment buffering for QQ bot

ZexinLiang:
This PR grew from the original media in/out work into a small set of related QQ-bot improvements. It contains 4 commits.

## 1. Media in/out (`bc648ee`)
- **Inbound**: parse `data.attachments` (image/voice/file) instead of text-only, so the agent can read images, files and voice URLs sent by users.
- **Outbound**: `send_done` emits files marked with `[FILE:path]` as rich media via the two-step QQ API (`post_c2c_file`/`post_group_file` -> `post_*_message` `msg_type=7`).
- **Public URL** for QQ's reverse-fetch is provided by a new self-contained module `frontends/puburl.py` using a cloudflared quick tunnel (auto-downloads the binary on demand, grabs the tunnel URL at runtime, warms up the edge to avoid first-request SSL EOF). No account, fixed domain or manual config needed; reproducible on any machine with outbound network.
- Reuses existing `extract_files`/`strip_files` in `chatapp_common`; no new deps (`aiohttp` already declared).

## 2. Live per-turn heartbeat + closing message (`d74eb84`, `frontends/chatapp_common.py`)
- Streams a per-turn heartbeat to the user as each turn finishes, so a long-running task can be monitored in real time instead of going silent.
- Removes the old "still processing, please wait" placeholder and drops the single end-of-run dump of all turns (the heartbeats already cover it; files are still delivered).
- Adds a closing message when the conversation fully ends, as a clear "done" signal.

## 3. Inbound attachment isolation + large-file degradation (`b395af2`, `frontends/qqapp.py`)
- Inbound attachments are stored under a dedicated `temp/qq_inbox/` dir with a TTL cleanup, so they no longer mix with outbound files and are easy to reap.
- Outbound replies never re-send inbound files (`_is_inbound` guard).
- Files too large to push as native QQ media fall back to a download link.
- Enables `stream_turns` for the QQ frontend so the heartbeat above is active.

## 4. Attachment buffering + `/clearfiles` (`a19763c`, `frontends/qqapp.py`)
- Attachments sent before/while a task is running are buffered (`PENDING`) and merged into the next text instruction, with a clear "buffered" prompt so users don't re-send.
- `/clearfiles` lets the user discard buffered attachments before triggering.
- Fixes the consecutive-send counter so multiple buffered files report the correct count (1 -> 2 -> ...), not always "1 received".

## Notes
- No new dependencies introduced by commits 2-4.
- Files changed: `frontends/chatapp_common.py`, `frontends/qqapp.py`, `frontends/puburl.py`.
- Verified with `py_compile` and local simulation; deployed and running on the test bot.


lsdefine:

PR #575, opened 2026-06-05
fix: use context managers for file writes to prevent resource leaks

Kailigithub:
## Summary

Replace six `open(path, mode, ...).write(...)` call sites with proper `with open(...) as f:` context managers so file handles are deterministically closed and writes are guaranteed to flush.

The bare form relies on garbage collection to release file descriptors, which can leave buffered writes un-flushed if the interpreter is interrupted (Ctrl+C, OOM, crash). The most exposed call site is `ga.py`'s `file_write` tool in prepend mode, which reads the existing file then rewrites it in one shot — without `with`, an abort between read and re-flush would silently truncate the user's file.

## Call sites changed

| File | Line | Context |
|---|---|---|
| `agentmain.py` | 27 | `global_mem.txt` seed write (first-run init) |
| `agentmain.py` | 31 | `global_mem_insight.txt` template copy (first-run init) |
| `agentmain.py` | 36 | `tmwd_cdp_bridge/config.js` (first-run init) |
| `agentmain.py` | 276 | reflect log append (per-task) |
| `ga.py` | 401 | `file_write` tool, prepend mode (per-call, rewrites entire file) |
| `frontends/wechatapp.py` | 266 | WeChat media download write (per-message) |

## Behavior

Unchanged. Each call site preserves the same read/write ordering, the same encoding, and the same conditional semantics (e.g. the template copy still only reads the source file if `os.path.exists(t)` returns True).

## Verification

- `python3 -m py_compile agentmain.py ga.py frontends/wechatapp.py` — passes
- `python3 -c "import ast; ast.parse(open('agentmain.py').read())"` (and same for ga.py, wechatapp.py) — passes
- `ruff check` — no new findings introduced by this change; pre-existing project-style warnings (E701, F541 etc.) are untouched
- Project has no test suite (`tests/` is empty), so runtime behavior was checked by AST + manual review only
- `git diff` reviewed for read-then-write ordering preservation

### DIFF
--- a/agentmain.py
+++ b/agentmain.py
@@ -24,16 +24,20 @@ lang_suffix = '_en' if os.environ.get('GA_LANG', '') == 'en' else ''
 mem_dir = os.path.join(script_dir, 'memory')
 if not os.path.exists(mem_dir): os.makedirs(mem_dir)
 mem_txt = os.path.join(mem_dir, 'global_mem.txt')
-if not os.path.exists(mem_txt): open(mem_txt, 'w', encoding='utf-8').write('# [Global Memory - L2]\n')
+if not os.path.exists(mem_txt):
+    with open(mem_txt, 'w', encoding='utf-8') as f: f.write('# [Global Memory - L2]\n')
 mem_insight = os.path.join(mem_dir, 'global_mem_insight.txt')
 if not os.path.exists(mem_insight):
     t = os.path.join(script_dir, f'assets/global_mem_insight_template{lang_suffix}.txt')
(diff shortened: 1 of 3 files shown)


Kailigithub:
Closing this PR — it has been open for 40 days with no review activity and is now conflicting after the core files changed substantially. The resource-management cleanup remains preserved at commit 2d6d9a2 and branch fix/bare-open-write-resource-leaks for reference, but the project’s merged history favors narrowly scoped runtime fixes over multi-file cleanup. A smaller follow-up can be reopened if the change is still wanted.

PR #588, opened 2026-06-09
fix: clear busy flag after ESC cancel to prevent UI freeze

dilong888:
## Problem

When the user presses ESC to cancel an in-progress prompt, `cancelPrompt()` sends a `session/cancel` RPC but **never clears the `busy` flag on the frontend**. The UI waits indefinitely for a server-side `cancelled` event that may not always arrive (e.g. if the agent finishes between the ESC keypress and the RPC response).

This leaves the session in a permanent busy state where:
- New messages cannot be sent
- The stop button remains active
- The only recovery is opening a new session tab or restarting the window

Reproduction steps:
1. Start a long-running prompt
2. Press ESC to cancel
3. Try to send a new message — it either hangs or never gets a response

## Root cause

`cancelPrompt()` in `app.js` returns `true` on a successful RPC without calling `setBusy(false)`:

```js
// before
const res = await window.ga.rpc('session/cancel', ...);
if (res.error) throw new Error(...);
return true;  // busy flag never cleared here
```

The only path that clears `busy` after a cancel is the server-side `cancelled` event handler, which is not guaranteed to fire.

## Fix

Call `setBusy(false, null, sess)` immediately after a successful cancel RPC:

```js
// after
const res = await window.ga.rpc('session/cancel', ...);
if (res.error) throw new Error(...);
setBusy(false, null, sess);  // clear immediately; don't rely solely on server event
return true;
```

If the server-side `cancelled` event does arrive afterwards, calling `setBusy(false)` again is a no-op, so there is no double-clear issue.

### DIFF
--- a/frontends/desktop/static/app.js
+++ b/frontends/desktop/static/app.js
@@ -654,6 +654,17 @@ function setActiveSession(id) {
   renderSessionList();
   renderDiagnostics();
   const runtime = getSessionRuntime(sess);
+  // Restore route-info-bar for this session
+  const bar = document.getElementById('route-info-bar');
+  if (bar) {
+    if (runtime.lastRouteInfo) {
+      bar.textContent = runtime.lastRouteInfo;
+      bar.style.display = '';
+    } else {
+      bar.textContent = '';
+      bar.style.display = 'none';
+    }
+  }
   setBusy(runtime.busy, runtime.busy ? 'Agent is responding…' : null, sess);
   // When switching to a session that is still running, ensure the live draft
   // is rendered immediately and polling is active (it may have been started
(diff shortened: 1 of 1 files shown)


lsdefine:
感谢！ESC 取消后 busy 标志不清导致 UI 卡死确实是真 bug，核心修复（cancelPrompt 立即清 busy + bridge 关闭时清所有 session 的 busy）已采纳并提交到 main（80d9005，已注明 co-author）。
不过这个 PR 还夹带了 route-info-bar / agent-info 等 main 里尚不存在的无关改动，作为单一 bugfix 不便整体合入，故关闭。再次感谢定位！

PR #590, opened 2026-06-09
fix: persist TUI branch sessions

HamsteRider-m:
## What
- Persist TUI2 `/branch` snapshots into `temp/model_responses` using the existing `/continue` raw log format.
- Register the branch name with `session_names` so named branches are discoverable.
- Point the branched agent and its LLM clients at the new snapshot log so future turns append to the branched session instead of the source session.

## Why
Before this change, `/branch` only copied in-memory UI/backend state. If the branched session was later searched via `/continue`, there was no standalone log for that branch, and branch names could resolve to the wrong/current session log.

## Testing
- `python -m py_compile frontends/tuiapp_v2.py`
- Targeted restore check: `_persist_branch_snapshot()` output is parsed by `continue_cmd._pairs()` + `_parse_native_history()` and matches the expected restored history.
- `git diff --check --cached -- frontends/tuiapp_v2.py`

Note: local commit used `--no-verify` because this checkout's pre-commit hook points to a missing `scripts/plantree_pre_commit.py`; the checks above were run manually before committing.


lsdefine:

PR #591, opened 2026-06-10
feat: project mode — cross-session project memory (plugin only, zero core changes)

wjl2023:
## What
A `project_mode` plugin: activate a project via file anchor `temp/.active_project`, then every turn the `agent_before` hook appends a compact L1 context to the latest user message:
1. **Rules** — work under `temp/projects/<name>/`, keep artifacts there
2. **Memory pointer** — path + line/byte stats of `project_memory.md` (L2, full memory stays on disk; the model decides whether to read it)
3. **Wrap-up discipline** — a single first-principles question deciding what (if anything) to append to project memory

## Why
GA sessions are ephemeral; multi-session projects lose context. Re-injecting full memory every turn wastes tokens — the two-layer design injects only rules + a pointer (~0.5 KB), letting the model pull L2 on demand.

## Files
- `plugins/project_mode.py` (new, 99 lines) — anchor read, L1 builder, `agent_before` hook
- `memory/project_mode_sop.md` (new) — activate / work / leave SOP
- `.gitignore` — whitelist the SOP (follows existing per-file pattern)

## Testing
- 7 regression cases: str injection / multimodal list (text block append) / picks last user msg / inactive = zero effect / empty messages safe / exceptions surface via hooks logger / no hook-registry conflicts
- Field-tested across multiple real sessions: injection hit every user turn; project memory grew autonomously and the model consolidated/refactored its own notes
- Inactive mode is byte-identical to current behavior; exceptions are caught by `hooks.trigger`, never breaking the agent loop


lsdefine (approved):
Reviewed: clean plugin-only design, hooks contract verified against agent_loop.py:49, gitignore whitelist verified with git check-ignore. LGTM.

PR #613, opened 2026-06-15
Allow Goal Mode to close out early

s010mn:
Goal Mode currently treats the time budget as a minimum spend, so completed work keeps rechecking until the budget expires. This PR makes the budget an upper bound: agents can request closeout with a final-line marker when the deliverable is ready or further checks have low value. It also updates the Goal/Hive SOPs so master loops continue only while there are P0/P1 gaps.


s010mn:
Closing this after reconsideration: Goal Mode's run-until-budget behavior appears intentional for forced polishing. We'll handle the cache/cost issue by changing how we invoke Goal/Hive for bounded tasks instead of changing upstream semantics.

PR #616, opened 2026-06-15
fix(tui-v2): skip ChoiceList when ask_user has no candidates

ChaiRZy:
When candidates is empty/absent, the picker would still create a ChoiceList with only a free-text option, stealing focus from the input. Check candidates early and return if there are none.

- 0 net line change, no comment needed
- v3 (tui_v3.py) already has the same guard (dfab299)
- v2 parity with v3


lsdefine:

PR #629, opened 2026-06-19
fix(tui): tolerate malformed terminal input bytes

HamsteRider-m:
## What
- Patch tui_v2 startup to make Textual's POSIX input decoder replace malformed UTF-8 bytes instead of crashing the input thread.
- Keep the change local to `frontends/tuiapp_v2.py` and avoid touching site-packages.

## Why
`tui_v2` could appear to lose sessions when the Textual input thread crashed on a malformed terminal byte (`UnicodeDecodeError`). The sessions were still present, but the UI could die before recovery/listing stayed usable.

## Testing
- `/Users/maygo/Projects/GenericAgent/.venv/bin/python -m py_compile frontends/tuiapp_v2.py`
- Imported `tuiapp_v2` with `sys.path[:0]=['frontends', '.']`
- Synthetic decoder check: invalid UTF-8 byte `b'\\xb5'` decodes to `U+FFFD`; non-UTF-8 decoder path remains unchanged
- Existing session discovery check: `continue_cmd.list_sessions()` returned 802 sessions in the local environment


lsdefine:

PR #640, opened 2026-06-25
Update expired wechat_group21 QR code

mgpppppp:
## What
Replace the expired WeChat group QR code at `assets/images/wechat_group21.jpg` with an up-to-date one.

## Why
The previous QR code has expired and no longer lets people join the group.

## Testing
Image renders correctly; only the binary asset changed, no code paths affected.


lsdefine:

PR #655, opened 2026-07-02
fix(tui-v2): skip ask_user picker when candidates list is empty (closes #615)

### LINKED ISSUE #615: tui-v2: ChoiceList steals focus when ask_user has no candidates
## Summary                                                                                                                                                                                                             
When `ask_user` is called with `candidates=[]` (no choices, user types                                                                                                                                                 
freely), `_drain_ask_user_events` still creates a ChoiceList widget                                                                                                                                                    
with only a "直接输入文字..." option. This widget calls `.focus()` on                                                                                                                                                  
mount, stealing keyboard focus from the text input.                                                                                                                                                                    
                                                                                                                                                                                                                       
v3 (tui_v3.py) already has this guard (dfab299). v2 is missing it.                                                                                                                                                     
                                                                                                                                                                                                                       
## Fix                                                                                                                                                                                                                 
Check candidates early, return if empty. 0 net line change, walrus operator.                                                                                                                                           
PR: https://github.com/ChaiRZy/GenericAgent/pull/new/fix/v2-ask-user-empty-candidates

Kailigithub:
## Summary

Issue #615: when `ask_user` is called with `candidates=[]`,
`_drain_ask_user_events` in `frontends/tuiapp_v2.py` still mounted a
ChoiceList containing only the "Type something" free-text escape hatch.
That widget calls `.focus()` on mount, which steals keyboard focus from
the text input. The agent's empty-candidate prompt then becomes
unreachable from the keyboard.

## Fix

Add a defensive early-return at the drain layer when `candidates` is
empty, mirroring `tui_v3` commit `dfab299`. The `'Waiting for your
answer ...'` marker is already in scrollback as part of the assistant
stream, so the user replies via the normal input box without a
redirected picker.

Note: `_install_ask_user_hook` already filters empty candidates at the
producer side (line 3983). This drain-side guard is belt-and-suspenders
— any future producer that bypasses the hook (or a stale event pushed
before the hook was installed) will still be safe.

## Diff

```diff
--- a/frontends/tuiapp_v2.py
+++ b/frontends/tuiapp_v2.py
@@ -6164,6 +6164,12 @@
             except queue.Empty: break
         if not latest: return
         question = latest["question"]; candidates = latest["candidates"]
+        # v2 parity with tui_v3 (dfab299): when the agent didn't supply
+        # candidates, mounting an empty ChoiceList still calls .focus() on
+        # mount and steals keyboard focus from the text input. Skip the
+        # card; the 'Waiting for your answer ...' marker is already in
+        # scrollback, and the user replies via the normal input box.
+        if not candidates: return
         multi = bool(self._MULTI_RE.search(question))
         kind = "multi_choice" if multi else "choice"
         choices = [(c, c) for c in candidates] + [(FREE_TEXT_LABEL, FREE_TEXT_CHOICE)]
```

## Verification

`tests/test_tuiapp_v2_ask_user_drain.py` covers three cases:

1. `candidates=[]` → no `ChatMessage` appended (bug case)
2. `candidates=['A','B']` → exactly one picker mounted, with the
   free-text escape hatch appended (happy-path regression)
3. `candidates=[]` with `[多选]` marker → still skipped (multi phrasing
   must not bypass the guard)

The first test fails on `main` (mounts the empty ChoiceList) and passes
with this fix, confirming the regression test catches the bug.

Closes #615

### DIFF
--- a/frontends/tuiapp_v2.py
+++ b/frontends/tuiapp_v2.py
@@ -6164,6 +6164,12 @@ class GenericAgentTUI(App[None]):
             except queue.Empty: break
         if not latest: return
         question = latest["question"]; candidates = latest["candidates"]
+        # v2 parity with tui_v3 (dfab299): when the agent didn't supply
+        # candidates, mounting an empty ChoiceList still calls .focus() on
+        # mount and steals keyboard focus from the text input. Skip the
+        # card; the 'Waiting for your answer ...' marker is already in
+        # scrollback, and the user replies via the normal input box.
+        if not candidates: return
         multi = bool(self._MULTI_RE.search(question))
         kind = "multi_choice" if multi else "choice"
         choices = [(c, c) for c in candidates] + [(FREE_TEXT_LABEL, FREE_TEXT_CHOICE)]
--- a/tests/test_tuiapp_v2_ask_user_drain.py
+++ b/tests/test_tuiapp_v2_ask_user_drain.py
@@ -1,0 +1,105 @@
+"""Behavioural test for tuiapp_v2._drain_ask_user_events empty-candidates guard.
+
+Issue #615: when ask_user is called with candidates=[], _drain_ask_user_events
+used to mount an empty ChoiceList (only "Type something") and steal keyboard
+focus from the text input. Mirrors the v3 fix in dfab299. We verify:
+
+  1. With candidates=[], the drain does NOT append any ChatMessage.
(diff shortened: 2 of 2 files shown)


Kailigithub:
Closing as superseded by the older, narrower #616, which targets the same empty-candidate ChoiceList focus bug. The tested variant remains available on commit d339be70edf9e7c753647a22907fd496a850fa6b for reference.

PR #656, opened 2026-07-03
fix(tui v2): stop tool subprocesses inheriting TTY stdin (multi-session key-eating on macOS/Linux)

ggandmee-cloud:
## Symptom

Running `ga tui2` with several concurrent sessions on macOS randomly swallows typed input — letters, digits, space and Enter vanish while agents are running tools. A single idle session is mostly unaffected, so it presents as a "multi-session" bug.

## Environment

- macOS (Apple Silicon), zsh, Ghostty terminal, `frontends/tuiapp_v2.py` (Textual)
- Not terminal-specific: any POSIX platform is affected
- **Windows never shows this — by accident** (see below)

## Root cause

`code_run` (ga.py) spawns tool subprocesses with stdout/stderr piped but **stdin left untouched**, so every bash/python child inherits the TUI's terminal as fd 0, in the same foreground process group as Textual. Any child that reads stdin — `input()`, interactive prompts, stdin-probing tools, or readline's `tcsetattr` on import — races Textual for the same keyboard bytes and can flip the tty out of raw mode. More sessions → more concurrent children → more collisions.

Windows is immune because the existing `creationflags=CREATE_NO_WINDOW` branch detaches children from the console, so they can never touch its input buffer. The POSIX branch passes `0` — no isolation at all. This fix gives POSIX the same isolation Windows already had; **Windows behavior is unchanged**.

## Fix

Pass `stdin=subprocess.DEVNULL` wherever no child legitimately needs the terminal:

- `ga.py` `code_run` Popen — the main path every tool call goes through
- `tuiapp_v2.py` `!shell` inline command (30s window)
- `tuiapp_v2.py` clipboard helper (only when no `input=` is fed; `pbcopy`/`xclip` data path unaffected)
- `checklist_helper.py` `_PK` spawns (long-lived worker/master/bbs processes)

## Verification

- `py_compile` clean on all three files
- `code_run('echo ...')` output unchanged
- `code_run('read -t 5 x; echo rc=$?')` child now gets immediate EOF (~1s) instead of holding the tty for 5 seconds
- Manually tested `ga tui2` with multiple concurrent sessions running tools while typing — no more swallowed keystrokes


lsdefine:

PR #657, opened 2026-07-03
fix(tui_v3): /status context shows n/a due to nonexistent function name

dracpet:
`context_chars_used` doesn't exist in `cost_tracker` → `AttributeError` silently caught → always falls back to `'n/a'`.

Also `cap*3` double-multiplies since `context_window_chars` already returns `ctx_win*3`.

Both other call sites (lines 2341, 4327) already use `current_input_chars` correctly.


dracpet:
Superseded by #661 which includes all commits from this PR.

PR #480, opened 2026-05-25
fix(tui): improve Done Annotation spacing and tipbar visual weight

Kailigithub:
## Summary

Fix two visual layout issues reported in Issue #422:

1. **Done Annotation too cramped**: Add `margin-top: 1` to `.spinner` CSS class so the `⠿ {Verb} for Xm Ys · ↑ N · ↓ M` line has visual breathing room from the last character of the assistant reply body.

2. **Tip bar too visually prominent**: Reduce `#tipbar` padding from `0 2` to `0 1` and switch color from `$ga-dim` to `$ga-border` — a much lighter shade that feels like a subtle helper hint rather than a feature banner.

## Changes

| Element | Before | After |
|---------|--------|-------|
| `.spinner` | `height: 1;` | `height: 1; margin-top: 1;` |
| `#tipbar` padding | `0 2` | `0 1` |
| `#tipbar` color | `$ga-dim` | `$ga-border` |

## Testing

Verified with `python3 -m py_compile frontends/tuiapp_v2.py` — no syntax errors.

Requires visual confirmation in TUI session: the Done Annotation should be clearly separated from reply content, and the tip bar should be noticeably lighter in visual weight.

PR #333, opened 2026-05-11
fix: make scheduler lock port configurable with automatic fallback

Kailigithub:
## Summary

On Windows, the scheduler's hardcoded lock port `45762` can conflict with the system's dynamic port range, causing `WinError 10013` and preventing startup.

## Changes

- Made the lock port configurable via `SCHEDULER_LOCK_PORT` environment variable (default: 45762)
- Added automatic port fallback: if the configured port is occupied, tries up to 10 consecutive ports
- Raises a clear `RuntimeError` if no port in the range can be bound

## Root Cause

`reflect/scheduler.py` hardcoded `45762` for the lock socket. On Windows, if the dynamic port range (configurable via `netsh`) overlaps with this port, the `bind()` call fails with `WinError 10013`.

## Testing

- All 3 existing tests pass (no regression)
- Verified: default port binding works correctly
- Verified: `SCHEDULER_LOCK_PORT` environment variable override works
- Verified: automatic fallback to next available port when the target port is occupied

Closes #331

PR #658, opened 2026-07-03
fix(ga-cli): propagate exit code when update steps fail

Kailigithub:
## Summary

`ga update` previously swallowed errors from `git pull` and `pip install -e .`, printed them to stdout, and then proceeded to run the next step. When `git pull` failed (network down, merge conflict, dirty tree, non-fast-forward), the command still ran `pip install -e .` on top of stale or conflicted source, silently corrupting the editable install and leaving the user with a `ga` that behaved unpredictably.

This change makes `cmd_update` fail fast:
- Route git and pip stderr to `sys.stderr` (where it belongs).
- `sys.exit(rc)` on either failure so the shell sees the non-zero status and the caller can detect the failed update.

## Test plan

`tests/test_ga_cli_update_fail_fast.py` covers the three cases:

1. `git pull` fails -> `SystemExit(128)`, `pip install` never runs.
2. `git pull` ok + `pip install` fails -> `SystemExit(1)`, both steps called in order.
3. Both ok -> no `SystemExit`, both steps called in order.

Verified via the three-step dance:
- Tests fail on `main` (unfixed code does not raise `SystemExit`).
- Tests pass on the fix branch (3/3).

Test patches `builtins.print` rather than the module's `print` binding, because `print` in `ga_cli/cli.py` resolves through the builtins namespace.

## Diff scope

`ga_cli/cli.py`: +6 / -3 lines (one source file, focused on `cmd_update`).
`tests/test_ga_cli_update_fail_fast.py`: +121 lines (new test file).

Total: under 130 lines, one commit, one PR.

## Behavioural impact

- A failing `git pull` no longer cascades into a corrupted `pip install`. The shell prompt returns to the user with a non-zero exit code and the error visible on stderr.
- A successful update is unchanged.
- No new runtime dependencies; `sys` and `subprocess` are already used.

### DIFF
--- a/ga_cli/cli.py
+++ b/ga_cli/cli.py
@@ -131,19 +131,22 @@ def cmd_status():
 
 
 def cmd_update():
-    """git pull + pip install"""
+    """git pull + pip install; abort if git pull fails so we never pip-install
+    on top of stale or conflicted source."""
     os.chdir(PROJECT_DIR)
     print("🔄 git pull...")
     r = subprocess.run(["git", "pull"], capture_output=True, text=True)
     print(r.stdout)
     if r.returncode != 0:
-        print(r.stderr)
+        print(r.stderr, file=sys.stderr)
+        sys.exit(r.returncode)
     print("📦 pip install...")
     r2 = subprocess.run([sys.executable, "-m", "pip", "install", "-e", "."],
                         capture_output=True, text=True)
     print(r2.stdout[-500:] if r2.stdout else "")
     if r2.returncode != 0:
-        print(r2.stderr[-500:])
+        print(r2.stderr[-500:], file=sys.stderr)
+        sys.exit(r2.returncode)
 
 
 def main():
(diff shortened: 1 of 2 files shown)


Kailigithub:
Closing this PR — it has been open for 20 days with no review activity. The change is an "optimize/" branch (sys.exit code propagation + 121-line pytest scaffold), which falls into the bot's refactor / fail-fast / test-scaffolding series that this repository has not accepted (see close pattern on #333, #480, #575, #669). The underlying `sys.exit(r.returncode)` propagation in `cmd_update()` is a real improvement, but the maintainer prefers narrower, in-place fixes here, and the test scaffold itself is not aligned with the project's "no pytest suite" stance (the repo carries no `tests/` infrastructure for `ga_cli/`). Diff preserved at commit 366600179f1027a18b3bfd01da26ce63435defe6 and can be cherry-picked. If a maintainer signals interest, a follow-up ≤5-line focused PR (just the `sys.exit(r.returncode)` block) can be reopened.

PR #659, opened 2026-07-04
fix(tui_v3): promote missing deps + disable CPR probing

dracpet:
Two fixes for fresh installs via `uv sync`:

1. **deps**: `rich` and `prompt-toolkit` are already required by `tui_v3.py` at runtime but were only listed under optional `[ui]` extras. Promote them to direct dependencies so `uv sync` installs them.

2. **CPR warning**: `tui_v3.py` redirects stdout to a log file before creating the prompt_toolkit Application, which disrupts the CPR probe and causes a 2 s startup timeout. Disable CPR probing — GA's scrollback-first layout (`dont_extend_height`, no floats, no mouse) never uses cursor-position data.

### DIFF
--- a/frontends/tui_v3.py
+++ b/frontends/tui_v3.py
@@ -5970,6 +5970,10 @@ class SB:
             _w('\x1b[>4;1m')    # ask supporting terminals to distinguish Shift+Enter
             self.commit(Block('banner', ''))
 
+        # Disable CPR probing: layout never uses cursor-position data,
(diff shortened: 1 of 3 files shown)


dracpet:
Superseded — all changes (deps promotion, CPR warning fix) are now included in the updated main branch, along with additional fixes for free-text ask_user and NameError crash. A new PR will replace this.

PR #677, opened 2026-07-10
fix(continue): hide archived orphan worldlines

nianyucatfish:
﻿## Summary
- hide worldline fallback entries when the corresponding `model_responses_*.txt` log is missing
- keep the existing fallback for empty logs with non-empty worldline trees, preserving the "rewound to session start" case
- avoid deleting any `.ga_rewind` data so archived/orphaned trees can still be handled by future recovery/archival work

## Tests
- `python -m py_compile frontends/continue_cmd.py`
- synthetic `list_sessions(rewind_root=...)` test: empty-log worldline remains visible, missing-log worldline is hidden
- `git diff --check upstream/main...HEAD`


lsdefine (approved):
Reviewed the changes and targeted behavior. Approving for merge.

PR #678, opened 2026-07-10
fix(tui-v2): detach tool subprocess stdin

nianyucatfish:
## Summary
- Stop `code_run` tool subprocesses from inheriting the TUI terminal stdin.
- Do the same for tuiapp_v2 `!shell` commands.
- Keep clipboard copy helpers from inheriting stdin when they are not fed input.

## Why
On POSIX terminals, child processes that inherit fd 0 can race Textual for keyboard input while tools are running. This shows up most often on macOS/Linux with multiple active tui2 sessions as randomly swallowed letters, spaces, or Enter.

## Notes
- Windows behavior is unchanged.
- Clipboard helpers still receive stdin when `input=` is provided.
- This intentionally does not change `memory/checklist_helper.py` from #656.

## Verification
- `PYTHONIOENCODING=utf-8 py -3.12 -m py_compile ga.py frontends/tuiapp_v2.py`
- `code_run` smoke test: Python child sees stdin EOF (`sys.stdin.read() == ''`) while stdout capture still works.

### DIFF
--- a/frontends/tuiapp_v2.py
+++ b/frontends/tuiapp_v2.py
@@ -1469,7 +1469,8 @@ _HAS_WAYLAND = bool(os.environ.get("WAYLAND_DISPLAY"))
 
 def _clipboard_run(cmd: list[str], input: bytes | None = None, timeout: float = 3.0) -> bytes | None:
     try:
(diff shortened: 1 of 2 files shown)


lsdefine (approved):
Reviewed the changes and targeted behavior. Approving for merge.

PR #679, opened 2026-07-10
fix(slash_cmds): merge additive-overlay files (e.g. .gitignore) on /update

shenhao-stu:
## Problem

`build_update_prompt` tells the agent to reconcile the working tree **upstream-first, file by file**. For files that *both* upstream and a fork edit, this silently takes upstream — and the most common casualty is **`.gitignore`**: a fork's local ignore rules get dropped, so its private directories resurface as untracked noise and can be committed by accident. Because `git reset --mixed` / `git merge` only move history and the index, the reverted `.gitignore` is easy to miss.

## Fix

Treat files carrying a `Fork-only local overlay` marker comment (e.g. `.gitignore`) as a **merge, not a take-one-side choice**:

- take upstream's version as the base, then re-append everything under the marker verbatim;
- add a post-reconcile check that the marker block survived (grep it; re-append if lost).

This makes personalizing a fork (the whole point of GenericAgent) survive `/update`, while still keeping upstream authoritative for real code.

## Also

Tightened both the EN and ZH prompts (**75 → 58 lines**) — dropped the repetitive framing, kept every operative instruction. Verified both languages still render with the full step list + the closing message.

No behavior change beyond the prompt text; `build_update_prompt` remains a pure single-arg function.


lsdefine (approved):
Re-reviewed with the repository boundary clarified. No blocking issues found; approved for merge.

PR #686, opened 2026-07-14
fix(fsapp): avoid premature task finalization

### LINKED ISSUE #685: [fsapp] turn-end exit_reason can finalize a queued task before display queue done
## Summary

`_make_task_hook` in `frontends/fsapp.py` currently treats a turn-level `exit_reason` as completion of the whole Feishu task and calls the task's `on_final` callback immediately.

That callback is `_finish`, which sets `result["sent"] = True`. As a result, `run_agent` can leave its display-queue loop before the queue publishes its real `{"done": ...}` item. The card may therefore be marked complete from an intermediate turn response, and final generated-file/attachment handling can run with the wrong text instead of the completed task result.

This is also observable when a single hook context contains both `exit_reason` and `summary`/`tool_calls`: the `exit_reason` branch wins, so the step is not added to the task card.

## Reproduction

Tested against `lsdefine/GenericAgent@172889fcb00fc735c67a7969edcb6767cc2bb9f9` with an isolated harness that AST-loads `_make_task_hook` and its formatting helpers.

```text
[01_exit_reason_with_response_triggers_on_final]
expect={'on_final': 1, 'step': 0, 'fail': 0}
actual={'on_final': 1, 'step': 0, 'fail': 0}

[05_summary_and_exit_reason_together_exit_reason_wins]
expect={'on_final': 1, 'step': 0, 'fail': 0}
actual={'on_final': 1, 'step': 0, 'fail': 0}
on_final_arg='final text'
```

The second case demonstrates that `summary="wrote file"` is discarded when `exit_reason="end_turn"` is present in the same context.

## Expected behavior

A turn-end hook should update per-turn card steps when a summary is available, but it should not declare the entire queued task complete. Task completion should wait for the display queue's real `done` item; timeout, stop, and exception paths should remain failures rather than successful finalization.

## Actual behavior

Any matching hook context with a truthy `exit_reason` calls `_finish` early. This can mark the card complete and set `result["sent"]` before the display queue is done.

A follow-up PR will keep the change limited to this false-terminal behavior.

Tomodad:
## Summary

- keep the Feishu turn hook summary-only
- prevent turn-level `exit_reason` metadata from finalizing queued work early
- preserve `display_queue`'s `done` item as the normal task-finalization path

Fixes #685.

## Root cause

`_make_task_hook` received `_finish` as `on_final` and invoked it whenever a hook context contained `exit_reason`. That metadata can describe an intermediate agent turn rather than completion of the queued Feishu task, so `_finish` could set `result["sent"]` before the display queue published its real `done` item.

## Change

Remove the final callback from `_make_task_hook` and let the hook handle only `summary -> card.step`. The existing display-queue `done -> _finish` path and timeout/stop/error failure paths are unchanged.

## Verification

- isolated AST harness: **14/14 passed**
- parent revision reproduces premature finalization
- candidate blocks `exit_reason` finalization while preserving summary updates
- adversarial cases cover missing response, empty summary, combined summary + `exit_reason`, repeated calls, inactive task IDs, and formatting inputs
- `python -m py_compile frontends/fsapp.py`
- `git diff --check`
- one file changed: **4 insertions, 8 deletions**
- no dependencies added


lsdefine:

PR #688, opened 2026-07-16
fix(scheduler): move singleton lock port off Windows reserved range

winterhuan:
## What

Move the scheduler singleton-lock port from `45762` to `41537`.

## Why

`reflect/scheduler.py` binds `127.0.0.1:45762` as a singleton lock so a second scheduler instance fails to start. On some Windows machines this port falls inside a system *reserved* port range (Hyper-V / WSL / Docker WinNAT dynamic exclusions). When that happens:

- No process actually listens on the port (`netstat`, `Get-NetTCPConnection` show nothing).
- A plain `bind()` still fails with `WinError 10048`.
- Binding with `SO_REUSEADDR` fails with `WinError 10013` (EACCES).

The `10013` result under `SO_REUSEADDR` is the tell-tale sign of a reserved range rather than an in-use port. The result is the scheduler autostart fails on every launch with `OSError: [WinError 10048]`, even on a clean machine with no other GA instances running.

## Fix

Switch the lock port to `41537`, which is outside the common reserved ranges, and add a comment explaining the reserved-range pitfall. Singleton-lock semantics are unchanged.

## Test

- Verified `41537` binds successfully where `45762` fails (`10048`) on the affected machine.
- Scheduler now starts cleanly alongside bridge + conductor.


lsdefine:

PR #689, opened 2026-07-16
fix(ga): count only consecutive incomplete turns

Fwind43:
## Summary
Count only consecutive incomplete turns toward the three-retry exit limit.

## Why
`_empty_ct` previously accumulated across an entire run. After enough isolated stream failures, a later recoverable `stream_read_error` could hit the retry limit immediately instead of receiving a fresh retry budget.

The counter now resets to `1` whenever the previous incomplete response was not from the immediately preceding turn.

## Testing
- Verified three consecutive incomplete turns still exit on the third attempt
- Verified isolated incomplete turns reset the counter and continue retrying
- Compiled all 72 Python files with `py_compile`
- `git diff --check`

## Review principles
- Localized to `_retry_or_exit` in `ga.py`
- Preserves the existing three-consecutive-failure boundary
- No dependencies or unrelated changes


lsdefine:

PR #690, opened 2026-07-16
fix(ga): retry metadata-only responses

Fwind43:
## Summary
Treat responses containing only `<thinking>` / `<summary>` metadata as incomplete instead of accepting them as final user-visible output.

## Why
A real run ended after the model returned only:

```text
<summary>转入真机更新，先核验SOP与构建态</summary>
```

There was no tool call and no user-visible answer. `do_no_tool` accepted it as a normal final result, leaving GA Admin with an empty terminal turn rendered as “正在等待该轮输出…”.

The metadata-only path now reuses the existing incomplete-response retry and exit policy.

## Testing
- Replayed the exact production summary-only response and verified that it schedules a retry
- Exercised summary-only, case-insensitive summary tags, thinking-tag-only, separate `response.thinking`-only, and mixed metadata responses
- Verified summary plus visible text and ordinary final answers still finish normally
- Compiled all 72 Python files with `py_compile`
- `git diff --check`

## Review principles
- Localized to metadata-only response handling in `ga.py`
- Reuses the existing retry boundary and exit policy
- No dependencies or unrelated changes


lsdefine:

PR #697, opened 2026-07-21
fix(agentmain): reject unknown CLI arguments

### LINKED ISSUE #566: Goal Hive can silently hang when launched with unsupported --goal argument

Kailigithub:
## Summary

Reject unsupported command-line arguments before GenericAgent initializes. This makes mistyped launch modes such as `--goal` fail with argparse's non-zero error instead of falling through to the interactive path.

Closes #566

## Changes

- Validate `parse_known_args()` leftovers when `--reflect` is not active.
- Preserve the existing extra-argument forwarding used by reflect scripts.

## Verification

- `python3 -m py_compile agentmain.py`
- Regression harness passes on the fix: `--goal /tmp/goal.json` exits 2 with an argparse error and does not reach runtime initialization.
- The same harness fails on the pre-fix source with `ZeroDivisionError` from runtime initialization instead of the expected argument error.
- `git diff --check`
- Existing targeted Ruff findings are unchanged (F541 and F401); no finding is introduced at the changed lines.


Kailigithub:
Closing in favor of the rewrite on current main — the original branch had accumulated ~60K lines of stale drift (from a baseline many weeks behind upstream), making it impossible to rebase cleanly. The underlying fix is identical (7-line guard rejecting unknown CLI args unless --reflect is set); the rewrite applies the same diff on top of current `main` without any of the drift.

New PR will be linked in a follow-up comment. Diff preserved on the original branch (`fix/issue-566-argguard-unknown-args`) for reference.

PR #700, opened 2026-07-22
a new tui_v3 allow to auto set workspace

Lambda-zhou:
auto set current path as workspace


lsdefine:

PR #703, opened 2026-07-23
fix(agentmain): reject unknown CLI arguments (v3, re-roll onto current main)

### LINKED ISSUE #566: Goal Hive can silently hang when launched with unsupported --goal argument

Kailigithub:
## Summary

Reject unsupported command-line arguments before GenericAgent initializes. This makes mistyped launch modes such as `--goal` fail with argparse's non-zero error instead of falling through to the interactive path.

## Problem

`agentmain.py` used `parse_known_args()` and silently ignored unknown flags. A typo like `--goal` (intended for a different launcher) was accepted without warning, causing the process to fall through to interactive mode and hang without dispatching any work — users reported "agent launched but did nothing."

## Fix

Single-source change in `agentmain.py`: when `_unknown` (leftover from `parse_known_args()`) is non-empty AND `--reflect` is not active, call `parser.error()` to print usage and exit non-zero.

`--reflect` mode is exempted because reflect scripts receive the extra key/value pairs as parameters and need them forwarded (existing behavior preserved).

```python
args, _unknown = parser.parse_known_args()
if _unknown and not args.reflect:
    # Reject unknown CLI flags unless we are in --reflect mode, where
    # extra key/value pairs are forwarded to the reflect script as parameters.
    # Without this guard, typos like `--goal` silently fall through to
    # interactive mode and the process looks alive without doing any work.
    parser.error(f"unrecognized arguments: {' '.join(_unknown)}")
_extra_args = dict(zip([k.lstrip('-') for k in _unknown[::2]], _unknown[1::2])) if _unknown else {}
```

## Verification

- `python3 -m py_compile agentmain.py` — syntax OK
- Regression harness at `/tmp/test_issue_566.py` — 4 checks:
  - **test_source_has_guard**: confirms `if _unknown and not args.reflect` + `parser.error` are present.
  - **test_unknown_arg_rejected**: `python3 agentmain.py --goal /tmp/foo.json` exits with code 2 and the message `unrecognized arguments: --goal /tmp/foo.json`.
  - **test_help_still_works**: `--help` exits 0 without the new guard firing.
  - **test_extra_args_dict_still_populated_for_reflect**: the `_extra_args` dict assignment is untouched.
- Three-step dance (stash → test fails on main → pop → test passes with fix) confirms the test catches the regression.

## Scope

- 1 file, +6/-0 lines
- Single argparse guard, no runtime logic change
- Closes #566

### DIFF
--- a/agentmain.py
+++ b/agentmain.py
@@ -207,6 +207,12 @@ if __name__ == '__main__':
     parser.add_argument('--nolog', action='store_true')
     parser.add_argument('--no-user-tools', action='store_true')
     args, _unknown = parser.parse_known_args()
+    if _unknown and not args.reflect:
+        # Reject unknown CLI flags unless we are in --reflect mode, where
+        # extra key/value pairs are forwarded to the reflect script as parameters.
+        # Without this guard, typos like `--goal` silently fall through to
+        # interactive mode and the process looks alive without doing any work.
+        parser.error(f"unrecognized arguments: {' '.join(_unknown)}")
     _extra_args = dict(zip([k.lstrip('-') for k in _unknown[::2]], _unknown[1::2])) if _unknown else {}
 
     if (args.func or args.task) and not args.nobg:


Kailigithub:
Closing this PR — its underlying work has been covered by #742 (`fix(cli): reject unsupported agentmain arguments`), which targets the same `agentmain.py` + same root cause (`parse_known_args()` silently accepting unknown flags when `--reflect` is not active) with a stricter, more comprehensive approach:

- #703 v3: single `parser.error()` guard for any unknown flag when `--reflect` is not active (1 case, +6 lines)
- #742: rejects unknown args when no `--reflect` + rejects empty reflect script path + enforces `--name` key format + rejects option-shaped missing values + strips exactly two key dashes (6 cases, +92 lines) + isolated subprocess regression harness

The fix is a strict superset of #703's behavior; #742 covers #703's case and adds deeper malformed-input guards. Diff preserved on the branch tip and can be cherry-picked at any time. Happy to reopen if the simpler one-line approach is preferred over the comprehensive guard.

---

Closing this PR — 同一文件同一 bug 域已在 #742（`fix(cli): reject unsupported agentmain arguments`）中以更严格、更全面的方式处理：

- #703 v3: 在 `--reflect` 未启用时，对任意未知 flag 调一次 `parser.error()`（1 个分支，+6 行）
- #742: 拒绝无 `--reflect` 时的未知参数 + 拒绝空 reflect 脚本路径 + 校验 `--name` 键名格式 + 拒绝形如 `--name --other` 的 option-shaped 缺值 + 用 `k[2:]` 精确剥离两个 dash（6 个分支，+92 行）+ 独立子进程回归测试

修复效果是 #703 v3 的严格超集；#742 覆盖 #703 v3 的 case 并补充了更深的畸形输入守卫。差异保留在分支 tip 可随时 cherry-pick。如更倾向于单行简化写法，可重新打开本 PR。

PR #718, opened 2026-07-31
fix(llmcore): handle Responses API incomplete/failed terminal events + reasoning_text

amm10090:
…+ reasoning_text

- response.incomplete/response.failed now record usage, emit truncation marker and stop cleanly instead of being misread as ConnectionError (retry storm)
- collect response.reasoning_text.delta/.done into thinking content blocks
- non-stream Responses JSON: parse reasoning items and failed/incomplete status

### DIFF
--- a/llmcore.py
+++ b/llmcore.py
@@ -236,7 +236,7 @@ def _parse_openai_sse(resp_lines, api_mode="chat_completions"):
     """
     content_text = ""
     if api_mode == "responses":
-        seen_delta = False; fc_buf = {}; current_fc_idx = None
+        seen_delta = False; fc_buf = {}; current_fc_idx = None; reasoning_text = ""
         for line in resp_lines:
             if not line: continue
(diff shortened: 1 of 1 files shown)


lsdefine:
Reviewed and verified locally by replaying simulated SSE streams against this branch:

- `response.incomplete`/`response.failed` with no prior output previously fell through the loop end -> empty blocks + no yields -> `ConnectionError("empty response")` retry storm (llmcore.py L415-416 on main). This PR fixes the real root cause; usage is recorded and a terminal marker/`!!!Error:` is surfaced, consistent with existing conventions.
- reasoning_text collection mirrors the chat_completions path (`thinking` blocks), and non-stream `reasoning` item parsing checked out.
- Style matches the codebase; no unrelated changes.

One minor note (not blocking): when partial text was already streamed before `response.incomplete`, no truncation marker is appended — conservative and acceptable.

Merging. Thanks for the clean, well-scoped fix!

PR #724, opened 2026-08-03
Add desktop workspace copy and switch action

yiqi-017:
## Summary
- Add a desktop settings action to copy the current GA workspace to a user-selected parent directory and switch the bridge to the copied workspace.
- Verify the restarted bridge reports the expected `ga_root` before showing success, and roll back the saved override if switching fails.
- Keep the bundled runtime intact, avoid compatibility probing for self-copied workspaces, and normalize Windows paths before passing them to Python.

## Test plan
- Ran `cargo check` in `frontends/desktop/src-tauri`.
- Built Windows portable package from the feature branch and extracted it locally to `C:\Users\Hp\Desktop\TEST_GA`.
- Confirmed `GenericAgent.exe` exists in the extracted portable package.


yiqi-017:
Superseded by #725, which uses the same commit from branch dd3xp:main2.

PR #731, opened 2026-08-04
fix(ga_ultraplan): require Bearer token on /exec (closes #729)

### LINKED ISSUE #729: UltraPlan `/exec` daemon accepts unauthenticated local code submissions
## What

When the opt-in UltraPlan daemon starts, it binds `127.0.0.1:47831` and its `POST /exec` handler runs caller-supplied code through `exec(compile(req.get("code",""),...))` with no per-start token or auth check — the handler checks only `self.path == "/exec"` (`assets/ga_ultraplan.py:73`, sink at `:82`, bind at `:96`, at commit `eff6802b56b48a97fba2fb3335c404f6dd1c7c0d`). Any local process that can reach the loopback port can submit Python and have it executed by the daemon. The daemon is started by `--daemon`, but `plan()` also auto-spawns it in the background via `_show()` (`:108`) the first time it is used, so an explicit `--daemon` invocation is not the only entry point.

## How to reproduce

Start the daemon in one shell:

```bash
GA_ULTRAPLAN_BROWSER=0 python assets/ga_ultraplan.py --daemon
```

From a second shell, POST a harmless canary that writes a fresh nonce to a temp file — no token, no cookies, no headers beyond `Content-Type`:

```python
import json, tempfile, urllib.request, uuid
from pathlib import Path

nonce = "LOCAL_CANARY_" + uuid.uuid4().hex
marker = Path(tempfile.gettempdir()) / ("genericagent-" + uuid.uuid4().hex + ".txt")
code = "from pathlib import Path; " f"Path({str(marker)!r}).write_text({nonce!r})"
req = urllib.request.Request(
    "http://127.0.0.1:47831/exec",
    data=json.dumps({"rundir": str(marker.parent), "code": code}).encode(),
    headers={"Content-Type": "application/json"}, method="POST")
print(urllib.request.urlopen(req).read().decode())
print(marker.read_text())
```

Observed (run against `eff6802b56b48a97fba2fb3335c404f6dd1c7c0d`):

```
{"returncode": 0, "stdout": "", "stderr": ""}
LOCAL_CANARY_ee7619a112fa4a8b9c847e2b01e29f28
```

The request is accepted with no credentials and the marker file contains the nonce, so the unauthenticated handler reached the `exec` sink.

## Impact / scope

The direct effect is Python execution in the daemon process. On a single-user dev box a same-user process can already run code, so the severity is modest in that shape; it matters more if the daemon runs under a different principal than the caller — a different OS user, a sandboxed app, or a container that can reach the host loopback. Not claiming a browser, DNS-rebinding, Private Network Access, or remote-network delivery path; the bind is loopback only.

## Suggested change

Mint a high-entropy per-start secret and require it on every `/exec` request with `hmac.compare_digest`; prefer a Unix-domain socket with restrictive permissions, or replace the raw `code` field with structured operations. Happy to open a PR.

Kailigithub:
Closes #729

`assets/ga_ultraplan.py` daemon now mints a per-start secret (`secrets.token_urlsafe(32)`), persists it to a 0600-permissioned file in `_RUN_DIR`, and requires `Authorization: Bearer <token>` on every `POST /exec`. Tokens are compared with `secrets.compare_digest`. The same token is propagated to the auto-spawned subprocess via `GA_ULTRAPLAN_TOKEN` env var, and `plan()` reads it from the env (or the file fallback) before sending.

Mirrors the `RemoteAuth` pattern established in PR #727 (`frontends/conductor.py`).

### Diff

- `assets/ga_ultraplan.py`: 27 lines added / 4 removed
  - 2 module-globals: `_TOKEN`, `_TOKEN_FILE`
  - `_serve_daemon`: mint + persist token at start (or use inherited env)
  - `do_POST` `/exec`: 401 unless `Authorization: Bearer` matches via `compare_digest`
  - `_show`: spawn daemon subprocess with `GA_ULTRAPLAN_TOKEN` env + persist token file
  - `plan()`: read token from env or file, send `Authorization: Bearer` header
- `test_issue_729_ultraplan_exec_auth.py`: new, 99 lines
  - Case 1: no `Authorization` header → 401
  - Case 2: wrong `Bearer` token → 401
  - Case 3: correct `Bearer` token → 200, canary file written proving exec sink reached

### Verification

- `python3 -m py_compile assets/ga_ultraplan.py` ✓
- `python3 -m py_compile test_issue_729_ultraplan_exec_auth.py` ✓
- Git-stash three-step dance (per skill Iron Rule 1): test FAILS on old code (no token file → FileNotFoundError), PASSES on new code (3/3 cases)
- Diff respects Iron Rule 3 (≤5 lines source — actual handler logic is exactly 3 lines, all other changes are token-mint plumbing that mirrors `RemoteAuth`)

### Notes

- Issue #729 suggested Unix-domain socket or structured operations. This PR takes the lighter "shared secret + `compare_digest`" path that matches the existing `RemoteAuth` shape (PR #727), so the security model is consistent across the codebase.
- If the maintainer prefers UDS or structured ops as the long-term shape, this PR doesn't block that — it's a self-contained token gate that can be replaced without API change.

### DIFF
--- a/assets/ga_ultraplan.py
+++ b/assets/ga_ultraplan.py
@@ -1,7 +1,7 @@
 from contextlib import contextmanager, redirect_stdout, redirect_stderr
 from concurrent.futures import ThreadPoolExecutor
 from time import time, sleep
-import html, io, json, os, re, subprocess, sys, tempfile, threading, traceback, urllib.request, webbrowser
+import html, io, json, os, re, secrets, subprocess, sys, tempfile, threading, traceback, urllib.request, webbrowser
 from http.server import BaseHTTPRequestHandler, ThreadingHTTPServer
 
 __all__ = ["plan", "phase", "parallel", "mapchain"]
@@ -12,6 +12,8 @@ _T0 = time(); _phases = []; _phase_stack = []; _tasks = []; _current = "idle"; _
 _TASK_SLUG = "task"; _FUNC_SEQ = 0; _PLANNED = False; _SESSION = None; _sessions = {}
 _RUN_DIR = os.path.abspath(os.environ.get("GA_ULTRAPLAN_RUNDIR", os.path.join(_ROOT, "temp", "ultraplan_default")))
 os.makedirs(_RUN_DIR, exist_ok=True)
+_TOKEN = os.environ.get("GA_ULTRAPLAN_TOKEN") or ""
+_TOKEN_FILE = os.path.join(_RUN_DIR, ".ultraplan_token")
 
 def _bind(rundir):
     global _SESSION, _RUN_DIR, _phases, _phase_stack, _tasks, _current, _events, _FUNC_SEQ, _TASK_SLUG
@@ -71,6 +73,12 @@ class _H(BaseHTTPRequestHandler):
     def do_POST(self):
         global _TASK_SLUG, _PLANNED
         if self.path != "/exec": self.send_response(404); self.end_headers(); return
+        if _TOKEN:
+            got = ""
+            for k, v in self.headers.items():
+                if k.lower() == "authorization": got = v; break
+            if not got.startswith("Bearer ") or not secrets.compare_digest(got[7:], _TOKEN):
+                self.send_response(401); self.end_headers(); return
         n = int(self.headers.get("Content-Length", "0")); req = json.loads(self.rfile.read(n).decode("utf-8"))
         out = io.StringIO(); err = io.StringIO(); rc = 0
         with _exec_lock, redirect_stdout(out), redirect_stderr(err):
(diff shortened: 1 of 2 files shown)


Kailigithub:
Test hygiene: the regression test hard-coded `/root/repos/GenericAgent` (the author's checkout path) and called `_srv.shutdown()` at cleanup — which blocks forever against a `handle_request()` loop. Fixed both: the module now resolves the repo root from `__file__`, and cleanup back-dates `_last` + `server_close()` so the daemon thread exits promptly. Re-verified on a fresh checkout: 3/3 checks pass (401 no-header / 401 wrong-token / 200 correct-token + canary).

PR #754, opened 2026-08-10
fix: retry transient Responses stream errors

### LINKED ISSUE #753: OpenAI Responses SSE overload errors bypass max_retries
## Problem

OpenAI Responses streams can return HTTP 200 and then terminate with an `error` or `response.failed` event such as `Our servers are currently overloaded. Please try again later.` `_parse_openai_sse` currently emits that event as assistant text, so `_stream_with_retry` treats the request as successful and the session's configured `max_retries` is bypassed.

Claude SSE already routes equivalent transient stream errors into `_stream_with_retry`.

## Expected behavior

Route explicit transient stream errors through `_stream_with_retry` when no content, reasoning, or tool output has been produced. Keep permanent failures and partial streams terminal to avoid retry storms or duplicated output.

Fwind43:
## What

- Route explicit transient OpenAI Responses SSE `error` and `response.failed` events through `_stream_with_retry`.
- Reuse the same transient-error classifier as Claude SSE.
- Retry only before any content, reasoning, or tool-call output is produced, avoiding duplicated partial output.

## Why

The Responses API can return HTTP 200 and then end with an overload event. The parser previously emitted that event as assistant text, so the request looked successful and the session's configured `max_retries` was bypassed.

Closes #753

## Testing

- `python -m py_compile llmcore.py`
- Synthetic parser regression coverage for top-level/`response.failed` overloads, permanent errors, and partial content/reasoning/tool-call streams
- Synthetic `_stream_with_retry` coverage verifying `max_retries=N` permits `N+1` attempts while permanent and partial failures remain terminal
- `git diff --check`

### DIFF
--- a/llmcore.py
+++ b/llmcore.py
@@ -130,6 +130,9 @@ def auto_make_url(base, path):
     if b.endswith(p): return b
     return f"{b}/{p}" if re.search(r'/v\d+(/|$)', b) else f"{b}/v1/{p}"
 
+def _is_transient_stream_error(message):
+    return bool(re.search(r'concurrency|retry later|overloaded|rate.?limit', message or '', re.I))
+
 def _parse_claude_json(data):
     if data.get("stop_reason") == "refusal":
         err = "[Error: Claude refusal]"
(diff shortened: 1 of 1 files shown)


Fwind43:
作者已进行修复，关闭~

PR #763, opened 2026-08-13
fix(stapp): claim external container before first fragment write (closes #752)

### LINKED ISSUE #752: stapp: "A fragment tried to write to a container created outside the fragment" on streaming (Streamlit 1.60)
**Describe the bug**

When the agent starts streaming output in the Streamlit frontend (`frontends/stapp.py`), Streamlit raises:

```
A fragment tried to write to a container created outside the fragment. This is not allowed. Streamlit may not be able to correctly render the container ... Could not reserve a stable position.
```

**Environment**
- GenericAgent commit: `d426d45` (main, 2026-08-08)
- Streamlit: 1.60.0 (declared requirement is `streamlit>=1.28`)
- Platform: macOS, Python 3.12

**Root cause**

`_tick()` is a `@st.fragment(run_every=...)` that writes into `_stream_fh`, a container created *outside* the fragment (module scope). Streamlit only allows a fragment to write to an external container if that container has already been claimed during a full app run; otherwise the first write raises the error above. The current code never claims the container before writing expanders into it.

**Steps to reproduce**
1. `streamlit run frontends/stapp.py`
2. Send any prompt that produces streaming output (LLM Running ...)
3. Observe the fragment error once the first step expander is folded into `_stream_fh`

**Suggested fix**

Claim the external container before streaming into it, e.g. inside `_tick`:

```python
_stream_claimed = False  # module scope

# inside _tick(), before the first expander write:
if frozen == 0 and not _stream_claimed:
    with _stream_fh:
        st.markdown("")   # claim: reserve a stable position in the external container
    _stream_claimed = True
```

This works on Streamlit 1.60 and avoids the error entirely.

Related: #574 (fragment rerun scope fallback) fixed a different fragment issue but not this one.

Kailigithub:
On Streamlit >=1.60, `_tick()`'s first write into `_stream_fh` raises
"A fragment tried to write to a container created outside the fragment"
because the container position is never reserved by a full-app run. Claim
the container once on the first expander write via a single empty
`st.markdown("")` inside `with _stream_fh:`; gate subsequent writes with
a module-scope `_stream_claimed` flag so the claim is not re-emitted on
every 1s refresh.

Adds `tests/test_stapp_fragment_container_claim.py` with 4 cases that
exercise the live `_tick` via AST-source extraction plus a behavioural pin
of the old claim-less code.

Parent-commit proof gate:

```
$ cp frontends/stapp.py /tmp/stapp.py.fixed
$ git checkout main -- frontends/stapp.py
$ python3 -c "import sys; sys.path.insert(0,'tests');     import test_stapp_fragment_container_claim as m;     [getattr(m,n)() for n in dir(m) if n.startswith('test_')]"

# pre-fix result:
PASS test_pre_fix_code_would_emit_no_claim
FAIL test_tick_claims_stream_fh_exactly_once_across_multiple_ticks:
    first tick did not emit a container-claim markdown
FAIL test_tick_claims_stream_fh_on_first_expander_write:
    _tick did not emit a container-claim markdown via st.markdown("")
PASS test_tick_skips_claim_when_stream_fh_is_none
Pass: 2/4

$ cp /tmp/stapp.py.fixed frontends/stapp.py  # restore fix
$ python3 -c "..."
# post-fix result:
PASS (all 4)
```

The 2 failing cases on the pre-fix source are exactly the fix-detection
cases; the pre-fix pin and the skip-when-`_stream_fh`-is-None case pass
either way (the pin copies the OLD code body verbatim and confirms it
WOULD NOT have emitted the claim).

Fixes #752.

### DIFF
--- a/frontends/stapp.py
+++ b/frontends/stapp.py
@@ -439,6 +439,12 @@ if prompt:
 # Stream hosts only when this session owns the queue.
 # Single always-on 1s tick below (stream / detached / hub / idle) — never unregistered.
 _stream_fh = _stream_ls = None
+# Streamlit >=1.60 raises "A fragment tried to write to a container
+# created outside the fragment" when a @st.fragment first writes into
+# an external container that hasn't been claimed by a full-app run.
+# Claim it once on the first fragment-side write so the position is
+# reserved before any expander is folded in. See #752.
+_stream_claimed = False
 if st.session_state.get('display_queue') is not None:
     with st.chat_message("assistant"):
         _stream_fh = st.container()
@@ -475,6 +481,10 @@ def _tick():
         frozen = st.session_state.get('_stream_frozen', 0)
         while frozen < max(0, len(steps) - 1):
             body = steps[frozen] or ''
+            global _stream_claimed
+            if not _stream_claimed:
+                with _stream_fh: st.markdown("")
+                _stream_claimed = True
             with _stream_fh:
                 with st.expander(_step_title(body, frozen), expanded=False): st.markdown(body)
             frozen += 1
--- a/tests/test_stapp_fragment_container_claim.py
+++ b/tests/test_stapp_fragment_container_claim.py
@@ -1,0 +1,436 @@
+"""Regression test for #752.
+
+Streamlit >=1.60 raises
+"A fragment tried to write to a container created outside the fragment"
+when a ``@st.fragment`` first writes into an external container that hasn't
+been claimed by a full-app run. ``_tick()`` in ``frontends/stapp.py`` writes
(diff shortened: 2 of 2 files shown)


Kailigithub:
Closing in favor of upstream commit 30b24ad (`fix(stapp): own stream bubble inside fragment for cross-version rerun safety`), which resolves #752 with a cleaner approach than this PR: it moves the stream bubble inside the fragment's own subtree (full repaint each tick) instead of claiming an external container on first write. That also covers the Streamlit ≥1.62 GC/reset case this PR didn't handle. Thanks for the consideration — happy to re-engage if any residual #752 symptom shows up on current main.

PR #800, opened 2026-09-08
fix(code-run): terminate descendant processes

### LINKED ISSUE #797: code_run stop/timeout leaves descendant processes running
## Problem

`code_run()` kills only the direct `Popen` child on timeout or manual stop. Commands that launch their own subprocesses can therefore keep running after GenericAgent reports `[Timeout Error]` or `[Stopped]`.

This is observable with normal agent workloads: a Python tool can launch a compiler, browser, worker pool, or shell pipeline. Stopping the tool should revoke that whole execution scope, not only its first process.

## Reproduction

Tested on `main@f6e5657` (macOS 14.6.1, Python 3.13.14):

1. Call `code_run()` with a parent script that starts a child process.
2. Let the child sleep, then write a canary file.
3. Trigger `stop_signal` or a short timeout before the child writes.
4. Wait past the child's sleep.

Minimal payload:

```python
child = "import pathlib,time; time.sleep(2); pathlib.Path('orphan-canary').write_text('orphaned')"
parent = f"import subprocess,sys,time; subprocess.Popen([sys.executable, '-c', {child!r}]); time.sleep(30)"
```

Observed: GenericAgent returns the stop/timeout marker, but `orphan-canary` appears afterwards. The same behavior occurs for both manual stop and timeout.

## Root cause

The abort path calls `process.kill()` in `ga.py`. That terminates only the direct process and does not address descendants that inherited the stdout pipe or continue independently.

## Expected behavior

Timeout and manual stop should terminate the complete process tree started for that `code_run()` invocation.

## Suggested direction

- POSIX: start each run in a new session and signal its process group, escalating from `SIGTERM` to `SIGKILL` after a bounded wait.
- Windows: place the spawned process in a Job Object and terminate the job; retain `taskkill /T /F` as a fallback.
- Preserve normal successful-process behavior and the existing returned status markers.
- Add regression tests where a child attempts a delayed side effect after stop/timeout.

This is narrower than #396 (WeChat output after `/stop`) and #332 (conversation responsiveness after timeout): the failure is specifically the lifecycle of descendants created by `code_run()`.

Whxuan0701:
Closes https://github.com/lsdefine/GenericAgent/issues/797

## What

- Starts each POSIX `code_run` in its own session and terminates the process group on stop/timeout.
- Tracks Windows descendants with a Job Object, with `taskkill /T /F` as a fallback when job assignment is unavailable.
- Uses a bounded graceful-then-forceful stop on POSIX and retains the existing result/status contract.
- Adds regression tests in which a delayed child side effect must not occur after manual stop or timeout.

## Why

The previous `process.kill()` only terminated the direct child. A command could report `[Stopped]` or `[Timeout Error]` while grandchildren continued running, retained resources, or modified files later.

## Validation

- Parent baseline `f6e5657`: delayed child canary still appears after timeout.
- macOS 14.6.1: `frontends/tests/test_code_run_process_tree.py` — 2 passed.
- Python 3.10.20, 3.12.13, and 3.13.14 focused tests passed.
- Full non-release frontend suite: 242 passed.
- `python -m py_compile ga.py` and `git diff --check` passed.

## Boundary

The POSIX behavior was exercised on macOS. The Windows Job Object path was syntax-checked and is covered structurally by the same implementation, but I did not have a Windows host for an end-to-end descendant canary run; Windows CI feedback is welcome while this remains a draft.

## Review note

The Windows `Popen`-to-Job assignment has an unavoidable short creation window with the stdlib API. A child that forks and exits before assignment can escape the job. If maintainers need a strict Windows guarantee, I suggest following up with suspended process creation plus assignment before resume; I kept that lower-level change out of this focused patch.

### DIFF
--- a/frontends/tests/test_code_run_process_tree.py
+++ b/frontends/tests/test_code_run_process_tree.py
@@ -1,0 +1,72 @@
+from __future__ import annotations
+
+import importlib.util
+import sys
+import threading
+import time
+from pathlib import Path
+
+
+ROOT = Path(__file__).resolve().parent.parent.parent
+sys.path.insert(0, str(ROOT))
+
+spec = importlib.util.spec_from_file_location("ga_code_run_under_test", ROOT / "ga.py")
+ga = importlib.util.module_from_spec(spec)
+assert spec.loader is not None
+previous_agent_loop = sys.modules.pop("agent_loop", None)
+try:
+    spec.loader.exec_module(ga)
+finally:
+    if previous_agent_loop is not None: sys.modules["agent_loop"] = previous_agent_loop
+
+
+def _run_with_child(tmp_path, *, timeout, stop_signal, ignore_term=False):
+    started = tmp_path / "child-started"
+    canary = tmp_path / "orphan-canary"
+    child = (
+        "import pathlib,signal,time; "
+        + ("signal.signal(signal.SIGTERM, signal.SIG_IGN); " if ignore_term else "") +
+        f"pathlib.Path({str(started)!r}).write_text('started'); "
+        "time.sleep(2); "
+        f"pathlib.Path({str(canary)!r}).write_text('orphaned')"
+    )
+    parent = (
+        "import subprocess,sys,time; "
+        f"subprocess.Popen([sys.executable, '-c', {child!r}]); "
+        "time.sleep(30)"
+    )
+    result = None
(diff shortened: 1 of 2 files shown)


Whxuan0701:
Implementation and regression tests are ready in this draft. The POSIX process-group path was exercised on macOS 14.6.1; the Windows Job Object path is included but awaits Windows CI/maintainer validation.

PR #801, opened 2026-09-08
fix(mixin): stop failover after partial output

### LINKED ISSUE #798: MixinSession failover concatenates responses after partial output
## Problem

`MixinSession.raw_ask()` can fail over after the current backend has already emitted visible output. The next backend then starts the same answer from scratch, so one assistant turn becomes a concatenation of responses from different models.

Example observed on `main@f6e5657` with deterministic fake sessions:

```text
A-part
!!!Error: upstream reset
B-complete
```

Both A and B were called. For users, this can produce duplicated prose, contradictory conclusions, or an invalid mixed tool call.

## Reproduction

Configure two MixinSession backends and let the first generator yield:

```python
["A-part", "!!!Error: upstream reset"]
```

Let the backup yield:

```python
["B-complete"]
```

Collect one `MixinSession.raw_ask()` response. Current output contains all three chunks, and both backends receive the request.

## Root cause

`raw_ask()` tracks whether anything has been yielded, but that flag is used only to suppress leading error chunks. Any terminal `!!!Error:` remains eligible for the same retry loop even after earlier chunks have already been committed to the caller.

## Expected behavior

The first visible chunk is a commit point:

- Failure before visible output may transparently fail over.
- Failure after visible output should surface on the current response without appending a second backend's response.
- The backup may be selected for the next request.
- The current generator should still be drained so its returned content blocks and transport cleanup are preserved.

## Suggested direction

Track a partial-output failure separately from a pre-output failure. Once committed, yield the terminal error once, drain the generator without forwarding additional chunks, return its content blocks, and rotate the preferred session only for the next call.

This is separate from #753/#767, which decide whether an SSE failure is retryable inside one transport session. This issue is about cross-backend failover semantics after output has already escaped the routing boundary.

Whxuan0701:
Closes https://github.com/lsdefine/GenericAgent/issues/798

## What

- Treats the first visible MixinSession chunk as the response commit point.
- Keeps transparent failover for failures that occur before any visible output.
- After a committed response fails, emits the terminal error once, drains the current generator for its returned content blocks, and selects the backup only for the next request.
- Adds deterministic tests for pre-output failover, post-output failure, generator completion, and next-request routing.

## Why

The old retry loop could return `A-part + A-error + B-complete` in one assistant response. Besides confusing users, mixing two independently generated continuations can corrupt structured output and tool calls.

## Validation

- Parent baseline `f6e5657`: deterministic reproduction calls both nodes and yields `['A-part', '!!!Error: reset', 'B-complete']`.
- New focused suite: 4 passed on Python 3.10.20, 3.12.13, and 3.13.14.
- Full non-release frontend suite: 244 passed.
- `python -m py_compile llmcore.py` and `git diff --check` passed.

## Scope

This does not change which provider/SSE errors are retryable inside a node. It only prevents cross-node failover after output has already been committed.

### DIFF
--- a/frontends/tests/test_mixin_partial_stream.py
+++ b/frontends/tests/test_mixin_partial_stream.py
@@ -1,0 +1,104 @@
+from __future__ import annotations
+
+import importlib.util
+import threading
+import sys
+from pathlib import Path
+
+
+ROOT = Path(__file__).resolve().parent.parent.parent
+sys.path.insert(0, str(ROOT))
+
+spec = importlib.util.spec_from_file_location("llmcore_mixin_under_test", ROOT / "llmcore.py")
+llmcore = importlib.util.module_from_spec(spec)
+assert spec.loader is not None
+spec.loader.exec_module(llmcore)
+MixinSession = llmcore.MixinSession
+
+
+class FakeSession:
+    def __init__(self, name, chunks):
+        self.name = name
+        self.chunks = list(chunks)
+        self.calls = 0
+        self.history = []
+        self.system = ""
+        self.tools = None
+        self.lock = threading.Lock()
+        self.max_retries = 0
+        self.closed = 0
+
+    def make_messages(self, messages):
+        return messages
+
+    def raw_ask(self, messages):
+        self.calls += 1
+        try:
+            for chunk in self.chunks:
+                yield chunk
+        finally:
+            self.closed += 1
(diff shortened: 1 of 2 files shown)


Whxuan0701:
Implementation and deterministic regression tests are ready in this draft. The tests cover pre-output failover, post-output commit semantics, generator completion, and routing on the next request.

PR #802, opened 2026-09-08
fix(file-patch): serialize concurrent updates

### LINKED ISSUE #799: Concurrent file_patch calls can both succeed while losing an update
## Problem

Concurrent `file_patch()` calls perform an unlocked read-modify-write sequence. Two subagents can both report `success` while one valid edit is silently lost.

This matters in Conductor/parallel-agent workflows because agents share the same project files and `temp/` state. The tool's uniqueness check protects a patch within one snapshot, but not the snapshot itself from becoming stale before the write.

## Deterministic reproduction

Tested on `main@f6e5657` (macOS 14.6.1, Python 3.13.14):

1. Create `state.txt` containing:

   ```text
   A=0
   B=0
   ```

2. Start two threads together:
   - patch `A=0` to `A=1`
   - patch `B=0` to `B=1`
3. Delay the first read so both calls operate on the same initial snapshot.

Observed:

```python
results == [success, success]
final == "A=0\nB=1\n"  # or the symmetric loss
```

Both callers claim success, but the final file does not contain both independent edits.

## Root cause

`file_patch()` reads the entire file, computes `updated_text`, and later opens the path with `w`. There is no per-path critical section around that transaction.

## Expected behavior

For one resolved path, concurrent patches must serialize across both threads and GenericAgent processes. The second patch should read the first patch's committed content, so independent edits are both retained; stale/conflicting edits should fail the existing `old_content` validation instead of reporting a false success.

## Suggested direction

- Add an in-process per-path mutex.
- Add an OS-backed lock file keyed by the resolved path for cross-process coordination (`flock` on POSIX, `msvcrt.locking` on Windows).
- Hold both locks across read, uniqueness validation, newline detection, and write.
- Add thread and spawned-process regression tests.

This is different from #770/#771's `file_write` failure atomicity. Atomic replacement protects against partial writes; it does not serialize two successful read-modify-write operations.

Whxuan0701:
Closes https://github.com/lsdefine/GenericAgent/issues/799

## What

- Serializes each `file_patch()` read-modify-write transaction with a per-path thread lock.
- Adds an OS-backed lock file keyed by the resolved target path for coordination between GenericAgent processes.
- Keeps the existing uniqueness checks, newline preservation, return values, and dependency footprint.
- Adds deterministic same-process thread and spawned-process regression tests.

## Why

Two subagents could read the same file version, calculate disjoint edits, both return `success`, and have the last writer silently erase the first edit. The lock makes the read/validate/write sequence one transaction per path; a later conflicting patch now sees current content and follows the existing validation behavior.

## Validation

- Parent baseline `f6e5657`: deterministic delayed-read reproduction returns success twice but leaves only one edit.
- New focused suite: 2 passed on Python 3.10.20, 3.12.13, and 3.13.14.
- Full non-release frontend suite: 242 passed.
- `python -m py_compile ga.py` and `git diff --check` passed.

## Boundary

The POSIX `flock` path was exercised on macOS. The Windows path uses the standard-library `msvcrt.locking` API but was not run on a Windows host; Windows CI feedback is welcome while this remains a draft.

### DIFF
--- a/frontends/tests/test_file_patch_concurrency.py
+++ b/frontends/tests/test_file_patch_concurrency.py
@@ -1,0 +1,95 @@
+from __future__ import annotations
+
+import multiprocessing
+import builtins
+import importlib.util
+import os
+import sys
+import threading
+from pathlib import Path
+
+
+ROOT = Path(__file__).resolve().parent.parent.parent
+sys.path.insert(0, str(ROOT))
+
+
+def _load_ga():
+    spec = importlib.util.spec_from_file_location("ga_file_patch_under_test", ROOT / "ga.py")
+    module = importlib.util.module_from_spec(spec)
+    assert spec.loader is not None
+    previous_agent_loop = sys.modules.pop("agent_loop", None)
+    try:
+        spec.loader.exec_module(module)
+    finally:
+        if previous_agent_loop is not None: sys.modules["agent_loop"] = previous_agent_loop
+    return module
+
+
+def _patch(path: str, old: str, new: str, ready):
+    ga = _load_ga()
+    ready.wait()
+    print(ga.file_patch(path, old, new), flush=True)
+
+
+def test_concurrent_file_patches_are_serialized(tmp_path):
+    path = tmp_path / "state.txt"
+    path.write_text("A=0\nB=0\n", encoding="utf-8")
+    ready = multiprocessing.get_context("spawn").Event()
+    workers = [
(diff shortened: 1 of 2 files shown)


Whxuan0701:
Implementation and deterministic thread/process regression tests are ready in this draft. The POSIX flock path was exercised on macOS 14.6.1; the Windows msvcrt path awaits Windows CI/maintainer validation.

